PR #33, opened 2021-11-15
Host operators

cliffburdick:
- BETA: Adding support for executing operators on the host
- Added support for compiling using a host compiler only (no nvcc)


cliffburdick:
Closes #33 #32

luitjens (approved):

PR #68, opened 2021-12-20
ND tensor

cliffburdick:
Added support for N-dimensional tensors. Previously MatX had a limit of rank 4 to make batching/indexing easier. This PR removes the restrictions for operators and libraries, and is only limited by the compiler/device memory.

Fixes #68


cliffburdick:
Fixes #68

PR #96, opened 2022-01-25
Added a Print() to print the whole tensor

ZJUGuoShuai:
Before, we have to call `.Print(0)`, `.Print(0, 0)`, `.Print(0, 0, 0)` or `.Print(0, 0, 0, 0)` to print the whole tensor (except 0-D tensors), which is not friendly. In addition, the [01_introduction.ipynb](https://github.com/NVIDIA/MatX/blob/main/docs_input/notebooks/01_introduction.ipynb) says that we could call `.Print()` on tensors to print the whole tensor.

So I added an alias to `.Print(0)`, `.Print(0, 0)`, `.Print(0, 0, 0)` or `.Print(0, 0, 0, 0)` by introducing an overloaded template function`Print(Args... dims)` that will be instantiated only when the arguments are empty and the rank of the tensor is greater than 0. This alias function actually constructs the `0, 0, ..., 0` arguments list and passes it to the original `Print(Args... dims)`.

The new interface has been tested on 0-D, 1-D, 2-D, 3-D and 4-D tensors.

### DIFF
--- a/include/matx_tensor.h
+++ b/include/matx_tensor.h
@@ -1553,9 +1553,11 @@ public:
    * @tparam Args Integral argument types
    * @param dims Number of values to print for each dimension
    */
-  template <typename ... Args, std::enable_if_t<((std::is_integral_v<Args>) && ...), bool> = true>
-  void Print(Args ...dims) const
-  {
+  template <typename... Args,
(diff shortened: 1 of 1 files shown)


cliffburdick:
Thanks @ZJUGuoShuai !

PR #182, opened 2022-05-25
Fix grid dimensions for transpose.

galv:
Includes a unit test that i used to reproduce my exact problem.

### DIFF
--- a/test/00_operators/OperatorTests.cu
+++ b/test/00_operators/OperatorTests.cu
@@ -750,6 +750,36 @@ TYPED_TEST(OperatorTestsComplex, ComplexTypeCompatibility)
   MATX_EXIT_HANDLER();
 }
 
+TYPED_TEST(OperatorTestsNumeric, ThreeDTranspose)
+{
+  MATX_ENTER_HANDLER();
+
+  index_t num_rows = 5998;
+  index_t num_cols = 64;
+
+  tensor_t<TypeParam, 3> t3 ({1, num_rows, num_cols});
+  tensor_t<TypeParam, 3> t3t({1, num_cols, num_rows});
+
+
+  for (index_t i = 0; i < num_rows; i++) {
+    for (index_t j = 0; j < num_cols; j++) {
+       t3(0, i, j) = static_cast<detail::value_promote_t<TypeParam>>(i * num_cols + j);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


test/00_operators/OperatorTests.cu
```diff
@@ -750,6 +750,36 @@ TYPED_TEST(OperatorTestsComplex, ComplexTypeCompatibility)
   MATX_EXIT_HANDLER();
 }
 
+TYPED_TEST(OperatorTestsNumeric, ThreeDTranspose)
```
cliffburdick, line 753: This should probably be in ViewTests.cu since it's not an operator test. 

Can you name it TransposeTest and we can just add other ranks later?

You can also remove my test in ViewTests.cu since the size was too small.

cliffburdick (changes requested):

test/00_operators/OperatorTests.cu
galv: Just to make sure about this: there are SquareCopyTranspose and NonSquareTranspose in OperatorTests.cu. So that's why I put this into OperatorTests.cu. Are you sure you want me to move it to ViewTests.cu? I'm happy to do so, of course.

cliffburdick: You're right -- I think this can just be added to those test cases as a 3D one inside the same test.

galv: @cliffburdick not quite sure I understand what you're saying. It sounds like you're suggested I should add more test code to the SquareCopyTranspose and NonSquareTranspose test cases to also test rank-3 tensors. I'm not sure how to do that easily in a generic way that is clear, since changing from rank 2 to rank 3 changes a bunch of the types, unless I just copy and pasted the body of my original test case.

Also, I would prefer to maintain a test case that caused the transpose errors I originally saw. 

I pushed a change that keeps my original test case in tact without touching those, while removing your original transpose on a 3D tensor. If you would like something different, can you just go ahead and make the change? The "Allow edits by maintainers" checkbox is ticked.

cliffburdick: That works too. I was just suggesting keeping a single set of transpose tests, one for square and one for rectangle, and we could add different ranks in there. But this works too. Thanks

cliffburdick (approved):

PR #183, opened 2022-05-26
Add unit test for mean + transpose calculation.

bpinzone-nvidia:
As it is the test `TYPED_TEST(ViewTestsFloatNonComplex, ViewTestMean)` inside `test/00_tensor/ViewTests.cu` fails.


test/00_tensor/ViewTests.cu
```diff
@@ -198,6 +203,85 @@ TYPED_TEST(ViewTestsIntegral, SliceStride)
   MATX_EXIT_HANDLER();
 }
 
+TYPED_TEST(ViewTestsFloatNonComplex, ViewTestMean)
```
bpinzone-nvidia, line 206: I don't think `ViewTestsFloatNonComplex` is what I should put here.

PR #192, opened 2022-06-08
Download RAPIDS.cmake only if it does not exist.

cwharris:
There's a bug where downstream repositories using a newer version of `RAPIDS.cmake` will have it overridden by MatX on subsequent builds. This prevents MatX from downloading `RAPIDS.cmake` if it already exists.


cliffburdick:
Thanks @cwharris!

PR #193, opened 2022-06-08
Added deriv/wrt operators.

luitjens:
The deriv operator will return the first derviative with respect to "X" of the underlying
operation.

The derivative of a constant (or tensor) will return 0.

The wrt operator signifies where the variable "X" is in the operation.

Fundamentally this operator has the derivative return 1 instead of 0 at
this point.  So wrt(tensor) will return 1 instead of 0.

The operator passed into wrt then becomes the points at which the
derivative is sampled.

wrt can appear multiple times in an operator but cannot be nested and
should have the same sample points passed into it.

Examples:

// df/dx sin(x)*sin(y), sampled at x
auto op = deriv(sin(wrt(x)) * sin(y));

// df/dy sin(x)*sin(y), sampled at y
auto op = deriv(sin(x) * sin(wrt(y)));

// df/dx sin(x)*sin(x), sampled at x
auto op = deriv(sin(wrt(x)) * sin(wrt(x)));

Limiations:
       deriv cannot be nested.  Only the first derivative can be
       computed.
       wrt cannot be nested.
       Not all operations support derivatives.

### DIFF
--- a/include/matx_scalar_ops.h
+++ b/include/matx_scalar_ops.h
@@ -32,8 +32,7 @@
 
 #pragma once
 
-#include <type_traits>
-
+#include "matx_scalar_diffs.h"
 namespace matx {
 namespace detail {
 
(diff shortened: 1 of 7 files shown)


luitjens:
I could use a good review of this.  In particular some of the scalar derivatives have not been tested.

include/matx_scalar_ops.h
```diff
@@ -133,11 +101,32 @@ template <typename T1, typename T2, typename T3, typename F> class TerOp {
...
 };
 
-MATX_UNARY_OP_GEN(ceil, Ceil);
-MATX_UNARY_OP_GEN(floor, Floor);
-MATX_UNARY_OP_GEN(round, Round);
-MATX_UNARY_OP_GEN(sqrt, Sqrt);
-MATX_UNARY_OP_GEN(exp, Exp);
+
+MATX_UNARY_OP_GEN(ceil, Ceil, zero);
+MATX_UNARY_OP_GEN(floor, Floor, zero);
```
cliffburdick, line 106: zero?

include/matx_scalar_ops.h
```diff
@@ -133,11 +101,32 @@ template <typename T1, typename T2, typename T3, typename F> class TerOp {
...
+MATX_UNARY_OP_GEN(round, Round, zero);
```
cliffburdick, line 107: zero?

include/matx_scalar_ops.h
```diff
@@ -133,11 +101,32 @@ template <typename T1, typename T2, typename T3, typename F> class TerOp {
...
+MATX_UNARY_OP_GEN(exp, Exp, exp);
+
+MATX_UNARY_OP_GEN(log10, Log10, dlog10);
+MATX_UNARY_OP_GEN(log2, Log2, dlog2);
+MATX_UNARY_OP_GEN(log, Log, dlog);
+MATX_UNARY_OP_GEN(abs, Abs, dabs);
+MATX_UNARY_OP_GEN(norm, Norm, none);
+
+// Trigonometric functions
+MATX_UNARY_OP_GEN(sin, Sin, cos);
```
cliffburdick, line 118: dsin?

include/matx_tensor_ops.h
```diff
@@ -1968,11 +2104,26 @@ auto __MATX_INLINE__ as_uint8(T t)
...
+    
+    template <typename... Is>
+    __MATX_DEVICE__ __MATX_HOST__ __MATX_INLINE__ auto deriv(Is... indices) const
+    {
+      //TODO is the compiler eliminating unecessary terms here?
+      auto v1 = get_value(in1_, indices...);         // value of left
+      auto v2 = get_value(in2_, indices...);         // value of right
+      auto d1 = get_deriv_value(in1_, indices...);    // diff of left
+      auto d2= get_deriv_value(in2_, indices...);     // diff or right
+#if 0
```
cliffburdick, line 2120: Remove?

include/matx_tensor_utils.h
```diff
@@ -253,23 +253,8 @@ namespace detail {
         return i(args...);
       }, sliced_tup);
     }
-
-    if constexpr (!(T::Rank() == sizeof...(Is))) {
```
cliffburdick, line 257: I'm pretty sure this was only fixed in 11.5, so we can't take it out if we want 11.4 support

test/00_diff/BasicDiffTests.cu
```diff
@@ -0,0 +1,204 @@
...
+     ASSERT_NEAR(t1o(i), cos(t1i(i))+sin(t1i(i)),.00001*scale);
+  }
+  
+  (t1o = deriv(sin(wrt(t1i))*cos(wrt(t1i)))).run(stream);
+  cudaStreamSynchronize(stream);
+  for (index_t i = 0; i <t1i.Size(0); i++) {
+     ASSERT_NEAR(t1o(i), cos(t1i(i))*cos(t1i(i))-sin(t1i(i))*sin(t1i(i)), .00001*scale);
+  }
+ 
+  if(!is_matx_half_v<TypeParam> ) {
```
cliffburdick, line 168: Why is half removed here? If it's the ambiguous cast can you cast it to a float if it's a half type?

include/matx_scalar_ops.h
```diff
@@ -133,11 +101,32 @@ template <typename T1, typename T2, typename T3, typename F> class TerOp {
   using scalar_type = std::invoke_result_t<decltype(op), T1, T2, T3>;
 };
 
-MATX_UNARY_OP_GEN(ceil, Ceil);
-MATX_UNARY_OP_GEN(floor, Floor);
-MATX_UNARY_OP_GEN(round, Round);
-MATX_UNARY_OP_GEN(sqrt, Sqrt);
-MATX_UNARY_OP_GEN(exp, Exp);
+
+MATX_UNARY_OP_GEN(ceil, Ceil, zero);
```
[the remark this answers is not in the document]
luitjens: I believe the derivative of a ceiling function is zero.  I'm not an expert.

include/matx_scalar_ops.h
luitjens: Ceiling and floor have zero for a derivative no?

include/matx_scalar_ops.h
luitjens: zero is the derivative according to my searching.

include/matx_tensor_ops.h
```diff
@@ -1924,7 +2042,23 @@ auto __MATX_INLINE__ as_uint8(T t)
...
+   
+    template <typename... Is>
+    __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ auto deriv(Is... indices) const 
+    { 
+      // chain rule 
+      auto gx = get_value(in1_, indices...);         // g(x),  compute g(x) at indices
+      auto gpx = get_deriv_value(in1_, indices...);   // g'(x), compute g'(x) at indices
+      auto fpgx = Op::deriv(gx);                      // f'(g(x)),  compute derivative at g(x)
+      auto retval = fpgx * gpx;                      // f'(g(x)) * g'(x),  chain rule
+#if 0
```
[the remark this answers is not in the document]
luitjens: left for debugging

test/00_diff/BasicDiffTests.cu
luitjens: half precision does not have enough dynamic range to resolve the solution in any meaningful range.  It compiles fine it just doesn't provide a useful answer.   I'll add a comment to clarify.

include/matx_scalar_ops.h
luitjens: fixed

include/matx_scalar_ops.h
cliffburdick: That makes sense. I guess it's something that wouldn't be used.

luitjens:
closing for now.  this was a proof of concept.

PR #205, opened 2022-07-08
Implementing remap operator.

luitjens:
This operator takes an operator/tensor
and an index array and logically gathers elements from that index array.
This allows the user to logically select/broadcast elements from a
dimension of the tensor.  This operator could then be stored in a memory
backed tensor to create a linear ordering of an operator to pass into a
library backed routine like matmul or fft.

### DIFF
--- a/include/matx_set.h
+++ b/include/matx_set.h
@@ -55,22 +55,20 @@ template <typename T, int RANK, typename Desc> class tensor_impl_t; ///< Tensor 
  * as the assignment operator.
  *
  * @tparam T
- *   Type of operator
- * @tparam RANK
- *   Rank of operator
+ *   Operator to use as output
  * @tparam Op
  *   Operator to use as input
  **/
-template <typename T, int RANK, typename Desc, typename Op>
-class set : public BaseOp<set<T, RANK, Desc, Op>> {
+template <typename T, typename Op>
+class set : public BaseOp<set<T, Op>> {
 private:
-  mutable tensor_impl_t<T, RANK, Desc> out_;
+  mutable typename base_type<T>::type out_;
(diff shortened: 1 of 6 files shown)


cliffburdick (approved):
Can you also add the function to docs_input?

include/matx_set.h
```diff
@@ -81,14 +79,15 @@ class set : public BaseOp<set<T, RANK, Desc, Op>> {
    * @param op
    *   Input operator
    */
-  inline set(tensor_impl_t<T, RANK, Desc> &out, const Op op) : out_(out), op_(op)
+  inline set(T &out, const Op op) : out_(out), op_(op)
   {
+    static_assert(is_matx_op_lhs<T>() == true, "Invalid operator on LHS of set/operator=");
```
cliffburdick, line 84: should this be named lvalue instead of lhs?

include/matx_tensor_ops.h
```diff
@@ -617,6 +617,64 @@ inline
...
+    constexpr __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ index_t Size(int dim) const
+    {
+      if(dim == dim_)
+        return idx_.Size(0);
+      else 
+        return op_.Size(dim);
+    }
+    
+    template<typename R> 
+    inline auto operator=(R rhs) {
```
cliffburdick, line 672: by reference?

include/matx_set.h
luitjens: changed

include/matx_tensor_ops.h
luitjens: made const reference

PR #237, opened 2022-08-08
Adding slice() operator.

luitjens:


include/matx_tensor_ops.h
```diff
@@ -773,6 +773,171 @@ auto __MATX_INLINE__ remap(Op t, Ind idx, Inds... inds)
...
+            d++;
+          }
+        }
+        MATX_ASSERT_STR(d==Rank(), matxInvalidDim, "SliceOp: Number of dimensions without matxDropDim must equal new rank.");
+      };
+
+      template <typename... Is>
+        __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ auto operator()(Is... indices) const 
+        {
+          static_assert(sizeof...(Is)==Rank());
```
cliffburdick, line 831: any chance we can have a version where there's no penalty for tensors?

luitjens: i'm not certain how that would look.  before we try to engineer around that we should first check if the overhead is significant or not.

include/matx_tensor_ops.h
```diff
@@ -773,6 +773,171 @@ auto __MATX_INLINE__ remap(Op t, Ind idx, Inds... inds)
...
+
+#pragma unroll 
+          for(int i = 0; i < T::Rank(); i++) {
+            ind[i] = starts_[i];
+          }
+
+#pragma unroll 
+          for(int i = 0; i < Rank(); i++) {
+            ind[dims_[i]] += inds[i] * strides_[i]; 
+          }
```
cliffburdick, line 846: is `inds` uninitialized here? maybe you mean `ind`?

include/matx_tensor_ops.h
cliffburdick: I'm not either, but every operator() access recomputing this would obviously be worse than what Slice does. not for this PR, but later it seems like we could figure it out.

include/matx_tensor_ops.h
cliffburdick: nevermind. I see it initialized with the array declaration.

include/matx_tensor_ops.h
luitjens: one option would be in the slice function to detect if op_ is a tensor and return op_.Slice(...); instead.  I'm still not convinced the overhead is sufficient for this extra path (x4) though.

cliffburdick: it's overhead that can be removed, plus it removes the need for Slice. it would be great if all those view functions were removed and still had no overhead for tensors.

PR #241, opened 2022-08-10
Fix cuTensorNet workspace size

### LINKED ISSUE #240: [BUG] `einsum` is using too much extra GPU memory
**Describe the bug**
`matx::cutensor::einsum` is using too much extra GPU memory. Is this a bug or not?

**To Reproduce**
1. Make 3 tensors `t1`, `t2` and `out`:
```cpp
auto t1 = matx::make_tensor<float, 2>({32, 10});
auto t2 = matx::make_tensor<float, 2>({32, 10});
auto out = matx::make_tensor<float, 1>({32});
```
2. Using gdb to stop here, I can use `nvidia-smi` to check memory usage:
image

3. Perform operation `einsum` on `t1` and `t2` to calculate `out`:
```cpp
matx::cutensor::einsum(out, "ij,ij->i", 0, t1, t2);
```

4. After that single call, check the memory:
image

**It just took >2GB memory**, is this normal?

**Expected behavior**
`matx::cutensor::einsum` may use less or no extra memory. BTW, if it does need extra memory, when are these memory `free`d?

**Code snippers**
```cpp
#include <matx.h>
#include <matx_einsum.h>

int main() {
    auto t1 = matx::make_tensor<float, 2>({32, 10});
    auto t2 = matx::make_tensor<float, 2>({32, 10});
    auto out = matx::make_tensor<float, 1>({32});

    matx::cutensor::einsum(out, "ij,ij->i", 0, t1, t2);

    cudaDeviceSynchronize();

    return 0;
}
```

**System details (please complete the following information):**
 - OS: Ubuntu 20.04
 - CUDA version: CUDA 11.4
 - g++ version: 9.3.0

**Additional context**
Tested on latest commit 128e05ac7258ee4362aa164ac07bfbaa157fc117.

leofang:
Close #240.

@ZJUGuoShuai any chance you can verify this patch? 🙂

### DIFF
--- a/include/matx_einsum.h
+++ b/include/matx_einsum.h
@@ -190,13 +190,13 @@ public:
 
     MATX_ASSERT_STR(workSize_ > requiredWorkspaceSize, matxOutOfMemory, "Not enough workspace memory is available.");
 
-    matxAlloc(&workspace_, workSize_, MATX_ASYNC_DEVICE_MEMORY, stream);
+    matxAlloc(&workspace_, requiredWorkspaceSize, MATX_ASYNC_DEVICE_MEMORY, stream);
 
     status = cutensornetWorkspaceSet(handle_,
                                           workDesc_,
                                           CUTENSORNET_MEMSPACE_DEVICE,
                                           workspace_,
-                                          workSize_);
+                                          requiredWorkspaceSize);
     MATX_ASSERT_STR(status == CUTENSORNET_STATUS_SUCCESS, matxcuTensorError,
       "Failed to set cuTENSOR workspace");     
 


leofang:
Note: this patch still keeps 2GB as the upper bound of the allowed workspace size. I think it's reasonable and can be relaxed in the future.

ZJUGuoShuai:
@leofang I tested this patch, now it only allocates ~500MB and that seems good to me :)

cliffburdick (approved):

leofang:
> Will check internally and see what we can do.

Forgot to update. Root cause identified.

PR #255, opened 2022-08-25
Adding unit tests for collapse with remap

luitjens:


test/00_operators/OperatorTests.cu
```diff
@@ -2618,6 +2618,90 @@ TEST(OperatorTests, Cast)
   MATX_EXIT_HANDLER();
 }
 
+TEST(OperatorTestsAdvanced, AdvancedOp)
```
cliffburdick, line 2621: Should this be a TYPED_TEST where it tests over many times?

cliffburdick (approved):

PR #262, opened 2022-09-07
Add streaming make_tensor APIs.

luitjens:


include/matx/core/make_tensor.h
```diff
@@ -131,20 +144,27 @@ auto make_tensor(T *ptr, const tensorShape_t<RANK> &shape, bool owning = false)
...
+ * @param stream cuda stream to allocate in (only applicable to async allocations)
  * @returns New tensor
  * 
  **/
 template <typename T, typename ShapeType, 
   std::enable_if_t< !is_matx_shape_v<ShapeType> && 
                     !is_matx_descriptor_v<ShapeType> && 
                     !std::is_array_v<typename remove_cvref<ShapeType>::type>, bool> = true>
-auto make_tensor(ShapeType &&shape) {
+auto make_tensor(ShapeType &&shape, matxMemorySpace_t space = MATX_MANAGED_MEMORY, cudaStream_t stream = 0) {
```
cliffburdick, line 157: I think we might want to change this to a class enum if we're exposing it more publicly now. Then this can't be accidentally cast from an integer.

cliffburdick (approved):

PR #267, opened 2022-09-08
Host unit tests

luitjens:


luitjens:
This should add some coverage for host compilation and running.  Note that we don't have the __CUDACC__ checks everywhere and will have to add them as we discover them in the future.

cliffburdick (approved):

PR #271, opened 2022-09-08
removing unused pybind call made before pybind initialize

tylera-nvidia:


tylera-nvidia:
@cliffburdick I found another unused call in the matlab, so removed that too

cliffburdick (approved):

PR #272, opened 2022-09-10
Support CubSortSegmented

tylera-nvidia:
Add Support for the newer CubSortSegmented over older RadixSort; Testing showed improved performance for SortSegmented through fewer kernel launches and higher occupancy. 

Outstanding actions:

- Waiting on CUB team to respond on question for 1D sorts
- Need to verify testing on older CUB.


cliffburdick:
@tylera-nvidia I assume this is still pending more benchmarking based on internal conversations?

tylera-nvidia:
@cliffburdick , yea I was waiting on Georgii to make sure I wasn't re-doing work. I'll get those results today and we can close this out

tylera-nvidia:
Results from RadixSort versus DeviceSegmented Sort:
image

Ultimately, 1Ds are always faster on Radix, 2Ds are faster on segment for small sizes, but Radix wins out once size increases. there is an obvious drop-off at 4096, so I think that's the right place to cut it off for now.

tylera-nvidia:
```
[==========] 1087 tests from 327 test suites ran. (237782 ms total)
[  PASSED  ] 1087 tests.

  YOU HAVE 44 DISABLED TESTS

docker@ipp1-1946:/home/tallen/scratch/MatX/build$ git rev-parse HEAD
4ebc15812b9dde78819c1bd5b9a0f108c010bbd2
docker@ipp1-1946:/home/tallen/scratch/MatX/build$ 
```

All tests passed at current hash

bench/00_transform/cub.cu
```diff
@@ -0,0 +1,80 @@
...
+#include <nvbench/nvbench.cuh>
+
+
+using namespace matx;
+
+//
+// types of data to run tests on
+//
+// using sort_types = nvbench::type_list<matxFp16Complex, cuda::std::complex<float>, cuda::std::complex<double>, float, double>;
+using sort_types = nvbench::type_list<double>;
```
cliffburdick, line 11: Can you add float too?

bench/00_transform/cub.cu
```diff
@@ -0,0 +1,80 @@
...
+  const int dataSize = static_cast<int>(state.get_int64("Tensor Size"));
+
+  auto sortedData = matx::make_tensor<ValueType>({dataSize});
+  auto randomData = matx::make_tensor<ValueType>({dataSize});
+
+  randomGenerator_t < ValueType > randData(dataSize, 0);
+
+  auto randTensorView = randData.template GetTensorView<sortedData.Rank()>(sortedData.Shape(), NORMAL);
+
+  (randomData = randTensorView).run();
```
cliffburdick, line 35: You probably need a prefetch here before running the benchmark since it's managed memory

bench/00_transform/cub.cu
```diff
@@ -0,0 +1,80 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+///
+///\brief benchmark for 1D Sorts
+///
+///\param param in
+///
+////////////////////////////////////////////////////////////////////////////////
+template <typename ValueType>
+void sort1d_range(
```
cliffburdick, line 21: You can probably remove "range" from the name since it's implied in the range that runs

bench/00_transform/cub.cu
```diff
@@ -0,0 +1,80 @@
...
+// types of data to run tests on
+//
+// using sort_types = nvbench::type_list<matxFp16Complex, cuda::std::complex<float>, cuda::std::complex<double>, float, double>;
+using sort_types = nvbench::type_list<double>;
+
+////////////////////////////////////////////////////////////////////////////////
+///
+///\brief benchmark for 1D Sorts
+///
+///\param param in
```
cliffburdick, line 17: The docstrings look incomplete. you can just remove them since these aren't parsed

include/matx/core/tensor_utils.h
```diff
@@ -56,23 +62,44 @@ namespace matx
...
+
+  /**
+   * @brief finds the size of the largest dim of the tensor
+   *core/tensor_utils.h
+   * @param op Operator
+   * @return size of largest dim
+   */
+  template <typename Op>
+  index_t LargestDimSize(const Op &op) {
+    index_t maxSize = 0;
```
cliffburdick, line 74: Slightly faster to set this to Size(0)  and iterate from 1..Rank()

include/matx/core/tensor_utils.h
```diff
@@ -56,23 +62,44 @@ namespace matx
...
+    for (int i = 0; i < op.Rank(); i++)
+    {
+      if( op.Size(i) > maxSize)
```
cliffburdick, line 77: See if 
```
maxSize = cuda::std::max(op.Size(i), maxSize);
```
I don't remember if it's in cuda::std yet

include/matx/transforms/cub.h
```diff
@@ -492,87 +494,269 @@ class matxCubPlan_t {
    *   Sort order (SORT_DIR_ASC or SORT_DIR_DESC)
    *
    */
-  inline void ExecSort(OutputTensor &a_out,
-                       const InputOperator &a,
-                       const cudaStream_t stream,
-                       const SortDirection_t dir = SORT_DIR_ASC)
-  {
+ inline void OptimizedExecSort(
```
cliffburdick, line 497: Can we just replace ExecSort if this is faster all-around?

cliffburdick (approved):

bench/00_transform/cub.cu
tylera-nvidia: they are incomplete, I'll fill them out with meaningful data.

What do you mean "they're not parsed"? that's valid doxygen syntax, do you mean the benchmark functions are not parsed by doxy?

include/matx/transforms/cub.h
tylera-nvidia: Optimized ExectSort is only faster in certain cases ( comment in ExectSort explaining cases)
`  // use optimized segmented sort if:
  //    - it is available (cub > 1.4)
  //    - it is greater than Rank 1 Tensor
  //    - it is < 8192 in all dims`
  
  Also OptimizedExecSort requires cub 1.4 or > which translates to CUDA 11.6 I believe? which I think is higher than our current CUDA requirement. 
  
  we could remove the ifdefs if we're ok moving the CUDA requirement up, but we'll still only want to use the Segmented calls for the conditions above.
  
  is the name misleading? I can rename to something like `ExecSegmentSort`

bench/00_transform/cub.cu
cliffburdick: we don't parse the benchmarks directly currently since they don't go anywhere

bench/00_transform/cub.cu
tylera-nvidia: updated for both float and double

bench/00_transform/cub.cu
tylera-nvidia: prefetched both

bench/00_transform/cub.cu
tylera-nvidia: ok cool. I went ahead and gave it meaningful content... the format pattern helps me with mental parsing so I'm ok with paying the dev maintenance to write the code.

include/matx/core/tensor_utils.h
tylera-nvidia: optimized to iterate from 1st element

include/matx/core/tensor_utils.h
tylera-nvidia: cuda:std::max isnt' there, but std::max is and is supported for these types, so changed to use that

tylera-nvidia:
Doubly-verified performance in 11.5, to ensure compilation checks were working:

`[----------] Global test environment tear-down
[==========] 1087 tests from 327 test suites ran. (242611 ms total)
[  PASSED  ] 1087 tests.

  YOU HAVE 44 DISABLED TESTS

docker@ipp1-1946:/home/tallen/scratch/MatX/build$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2021 NVIDIA Corporation
Built on Thu_Nov_18_09:45:30_PST_2021
Cuda compilation tools, release 11.5, V11.5.119
Build cuda_11.5.r11.5/compiler.30672275_0
docker@ipp1-1946:/home/tallen/scratch/MatX/build$ git rev-parse HEAD
[b0a8b832f11b4306b4ada265b56a8cfdf1a1a9eb]
docker@ipp1-1946:/home/tallen/scratch/MatX/build$ `

PR #280, opened 2022-09-29
Remove Deprecated Shape and add metadata to Print

tylera-nvidia:
Remove the deprecated shape class, updating the RNG class to use newer array types. 
Also added info to the print utility to include metadata about the type of class printing, it's Size/Rank/Stride, as appropriate per input class.


include/matx/core/tensor_utils.h
```diff
@@ -440,6 +440,40 @@ namespace detail {
...
 
+  /**
+   * convert Type to string
+   *
+   * function convert a tensor type to a string
+   *
+   */
+  template <typename T> static std::string GetTensorType()
+  {
+    if constexpr (std::is_same_v<T, int32_t>)
```
cliffburdick, line 451: I would change these to be the real C++ datatypes. The ones from pybind.h were specific to numpy, but it's useful in c++ to see things like int32_t instead of i4.

include/matx/core/tensor_utils.h
```diff
@@ -440,6 +440,40 @@ namespace detail {
...
+                  std::is_same_v<T, std::complex<double>>) {
+      return "c8";
+    }
+    if constexpr (std::is_same_v<T, cuda::std::complex<float>> ||
+                  std::is_same_v<T, std::complex<float>> ||
+                  is_complex_half_v<T>) {
+      return "c4";
+    }
+
+    return "";
```
cliffburdick, line 473: Maybe do "Unknown" here?

include/matx/generators/random.h
```diff
@@ -178,38 +177,17 @@ template <typename T> class randomGenerator_t {
...
-   */
-  template <int RANK>
-  inline auto GetTensorView(const tensorShape_t<RANK> shape,
-                            Distribution_t dist, T alpha = 1, T beta = 0)
-  {
-    return randomTensorView_t<T, RANK>(shape, states_, dist, alpha, beta);
-  }
-
   /**
    * Get a tensor view of the random numbers
```
cliffburdick, line 192: This looks like it should be `shape` still

cliffburdick (approved):

include/matx/core/tensor_utils.h
tylera-nvidia: ok I will update. I didn't know if this was shorthand people preferred that just happened to line up with python types. I'll make them the real types

include/matx/core/tensor_utils.h
tylera-nvidia: updated for all C++ types supported by MatX and added unknown

include/matx/generators/random.h
tylera-nvidia: updated this comment and added a function header to the overloaded one below

PR #281, opened 2022-09-30
Adding throw to file reading

tylera-nvidia:
Adding throw to file reading


include/matx/core/file_io.h
```diff
@@ -134,6 +135,12 @@ void ReadCSV(TensorType &t, const std::string fname,
                "CSV reading limited to tensors of rank 1 and 2");
   }
 
+  struct stat buffer;  
+  if (stat(fname.c_str(), &buffer) != 0) {
```
cliffburdick, line 139: can you use https://en.cppreference.com/w/cpp/filesystem/exists instead to be more c++ish?

cliffburdick (approved):

PR #283, opened 2022-10-06
Adding str() function to generators and operators

luitjens:


cliffburdick (commented):
it's probably too late now, but if we make every variable constexpr static char* we can still construct the strings with std:: string, but it makes it possible to do compile-time generation later.

luitjens:
> it's probably too late now, but if we make every variable constexpr static char* we can still construct the strings with std:: string, but it makes it possible to do compile-time generation later.

I first started with compile time but due to limitations with std::string I couldn't get it to work.  We can always revisit later.  Right now I think trying to get it to be compile time would be more hassle than it is worth.  This has zero impact on production runs as getstring won't be called.  It likely has very minimal impact on profiling runs.

cliffburdick (approved):

PR #304, opened 2022-10-19
Meshgrid API Changes

luitjens:
Update Meshgrid APIs to work with ND tensors and return operators for the output tensors.

### DIFF
--- a/include/matx/generators/meshgrid.h
+++ b/include/matx/generators/meshgrid.h
@@ -31,89 +31,87 @@
 /////////////////////////////////////////////////////////////////////////////////
 
 #pragma once
-
+#include "matx/operators/permute.h"
 
 namespace matx
 {
+
+
   namespace detail {
-    template <typename T> class Meshgrid_X {
-      private:
-        std::array<T, 3> x_;
-        std::array<T, 3> y_;
 
-      public:
-        // dummy type to signal this is a matxop
-        using matxop = bool;
-        using scalar_type = T;
+    template <typename T1, int RANK, int AXIS> 
+      class MeshGridOp {
+        private:
+          T1 t1_;
+          std::array<index_t, RANK> shape_;
 
-        __MATX_INLINE__ std::string str() { return "meshgridx"; }	
+        public:
+          using matxop = bool;
+          using scalar_type = typename T1::scalar_type;
+          //typedef typename T1::scalar_type scalar_type;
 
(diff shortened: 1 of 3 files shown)


include/matx/generators/meshgrid.h
```diff
@@ -31,89 +31,89 @@
...
 
-        Meshgrid_X(std::array<T, 3> x, std::array<T, 3> y) : x_(x), y_(y) {}
+          __MATX_INLINE__ std::string str() { return "meshgrid"; }
 
-        inline __MATX_DEVICE__ T operator()(index_t i, index_t j) const
-        {
-          return x_[0] + j * (x_[1] - x_[0]) / (x_[2] - 1);
-        }
+          __MATX_INLINE__ MeshGridOp(T1 t1, std::array<index_t, RANK> shape) : t1_(t1), shape_(shape) {
+            static_assert(shape.size() == RANK );
```
cliffburdick, line 56: Maybe add a second argument to each of these saying why the assert failed

include/matx/generators/meshgrid.h
```diff
@@ -31,89 +31,89 @@
...
+      };
+
+    template <typename... Ts, int... I>
+      __MATX_INLINE__  auto meshgrid_impl(std::integer_sequence<int, I...>, Ts&... ts) {
+        auto constexpr RANK = (int)sizeof...(Ts);
+
+        // construct shape from size of each rank1 tensor
+        std::array<index_t, RANK> shape{ts.Size(0)...};
+
+        // Python XY indexing is reverse from natural tensor indexing.  We premute here to match
```
cliffburdick, line 85: "permute"

include/matx/generators/meshgrid.h
```diff
@@ -31,89 +31,89 @@
...
+        int perm[RANK];
+        for(int i=0; i < RANK; i++) {
+          perm[RANK-i-1] = i;
         }
-        static inline constexpr __MATX_HOST__ __MATX_DEVICE__ int32_t Rank() { return 2; }
-    };
-  }
+
+        // TODO transpose is not correct,  we need to reverse dimensions
```
cliffburdick, line 91: Is this comment old?

cliffburdick (approved):

include/matx/generators/meshgrid.h
luitjens: a user will never hit this assert since they use a different api.  it was more for matx developers.

PR #308, opened 2022-10-21
Adding reduce apis that take reduction dims

luitjens:


include/matx/operators/permute.h
```diff
@@ -137,10 +146,27 @@ namespace matx
...
+   * This operator can appear as an rvalue or lvalue. 
+   *
+   * @tparam T Input operator/tensor type
+   * @param op Input operator
+   * @param dims the reordered dimensions of the operator.
+   * @return permuted operator
+   */
+  template <typename T>
+    __MATX_INLINE__ auto permute( const T &op, 
+        const std::array<int32_t, T::Rank()> &dims) {
```
cliffburdick, line 167: instead of having two identical functions in the class the C array version should call detail::toarray() and just convert and can this function.

luitjens: fixed

include/matx/transforms/reduce.h
```diff
@@ -1269,6 +1268,53 @@ void __MATX_INLINE__ sum(OutType dest, const InType &in, cudaStream_t stream = 0
...
+    if(!visited[i]) {
+      perm[j++] = i;
+    }
+  }
+
+  return perm;
+}
+
+template <typename OutType, typename InType>
+void __MATX_INLINE__ sum(OutType dest, const InType &in, std::initializer_list<int> dims, cudaStream_t stream = 0)
```
cliffburdick, line 1302: should use `const index_t (&arr)[R]` instead I think with R as a template value type?

include/matx/transforms/reduce.h
```diff
@@ -1269,6 +1268,53 @@ void __MATX_INLINE__ sum(OutType dest, const InType &in, cudaStream_t stream = 0
 #endif  
 }
 
+template <int RANK, class T>
+auto __MATX_INLINE__ getPermuteDims( T &dims ) {
```
cliffburdick, line 1272: const?

include/matx/transforms/reduce.h
```diff
@@ -1269,6 +1268,53 @@ void __MATX_INLINE__ sum(OutType dest, const InType &in, cudaStream_t stream = 0
...
+  std::array<int, RANK> perm;
+  std::array<bool, RANK> visited;
+
+  visited.fill(false);
+  
+  // construct permutation array by moving fastest changing dims to end
+  int j = RANK-1;
+  for(auto it = rbegin(dims); it != rend(dims); it++) {
```
cliffburdick, line 1280: `crbegin`/`crend`?

test/00_operators/ReductionTests.cu
```diff
@@ -225,6 +225,44 @@ TYPED_TEST(ReductionTestsNumericNoHalf, Sum)
...
     }
   }
+  
+  {
+    tensor_t<TypeParam, 2> t2a({30, 40});
+    tensor_t<TypeParam, 2> t2b({30, 40});
+    
+    auto t4 = ones<TypeParam>({30, 40, 50, 60});
+
+    sum(t2a, t4, 0);
```
cliffburdick, line 235: You can remove the stream if you want to make it a little more readable when we point people here

test/00_operators/ReductionTests.cu
```diff
@@ -225,6 +225,44 @@ TYPED_TEST(ReductionTestsNumericNoHalf, Sum)
...
+    auto t4 = ones<TypeParam>({30, 40, 50, 60});
+
+    sum(t2a, permute(t4,{2,3,0,1}), 0);
+    sum(t2b, t4, {0,1}, 0);
+
+    cudaStreamSynchronize(0);
+    for (index_t i = 0; i < t2a.Size(0); i++) {
+      for (index_t j = 0; j < t2a.Size(1); j++) {
+        EXPECT_TRUE(MatXUtils::MatXTypeCompare(
+            t2a(i, j), t2b(i,j)));
```
cliffburdick, line 261: Have you verified these match exactly with what Python does? It would be useful to have all reduction tests match Python, but that's a bigger task.

cliffburdick:
Fixes #308

cliffburdick (approved):

tylera-nvidia (approved):

include/matx/transforms/reduce.h
luitjens: I'm not sure I follow.  Could you explain more?

test/00_operators/ReductionTests.cu
luitjens: yes

include/matx/transforms/reduce.h
cliffburdick: Instead of `initializer_list`, which has other problems, we should use a C-array so it can use the `{}` syntax but also allow arrays to be passed in like the other functions.

luitjens: we have the array APIs also.  Maybe initializer_list isn't correct.  Perhaps a vector would be better?

luitjens: now using vector.  if acceptable please resolve this conversation.

include/matx/transforms/reduce.h
luitjens: fixed

include/matx/transforms/reduce.h
cliffburdick: I think I might be missing context here, but all of these APIs should be using `const index_t (&arr)[R]` and not vector or initializer_list. vectors have extra overhead that isn't needed (dynamic allocations), and initializer_lists have their own set of problems. I think you can just replace all the instances of vector with C arrays, ie:

```
template <typename OutType, typename InType>
void __MATX_INLINE__ stdd(OutType dest, const InType &in, std::vector<int> dims, cudaStream_t stream = 0)
```
becomes
```
template <typename OutType, typename InType, int R>
void __MATX_INLINE__ stdd(OutType dest, const InType &in, const int (&dims)[R], cudaStream_t stream = 0)
```

luitjens: not for this parameter as the size of R is not necessarily known at compile time.  We know it is <= dims.  We could add another template parameter as you do above and force it throughout the APIs if we deem that better but I think it is honestly unnecessary.

include/matx/transforms/reduce.h
luitjens: fixed

include/matx/transforms/reduce.h
luitjens: fixed

PR #309, opened 2022-10-25
Median nd

luitjens:


include/matx/transforms/reduce.h
```diff
@@ -1229,52 +1229,77 @@ void __MATX_INLINE__ median(OutType dest,
...
+
+    constexpr int batch_dims = OutType::Rank() - 1;
+    constexpr int red_dims = OutType::Rank() - TensorInType::Rank();
+
+    if constexpr (batch_dims > 1) {
+      // collapse batch dimensions to a single dimension
+      auto oop = lcollapse<batch_dims>(dest);
+      auto iop = lcollapse<batch_dims>(in);
+
+      static_assert(oop.Rank() <= 1);
```
cliffburdick, line 1288: Maybe add an error string on all these static asserts to tell them what's wrong?

cliffburdick (approved):

include/matx/transforms/reduce.h
luitjens: these are actually just sanity check asserts that should be impossible to kick up.  They are there for people who are modifying this specific code (aka validating my logic was correct).

PR #313, opened 2022-10-26
Change Semantics of collapse APIs

luitjens:


luitjens:
Note to update from old semantics to current semantics just add one to the template parameter.

cliffburdick (approved):

tylera-nvidia (approved):

PR #319, opened 2022-11-04
Support axis selection and operators in FFTs

luitjens:
PR working towards supporting dimensions selection and operators support.  Not ready for review.

### DIFF
--- a/include/matx/core/tensor.h
+++ b/include/matx/core/tensor.h
@@ -217,6 +217,16 @@ public:
       : detail::tensor_impl_t<T, RANK, Desc>{rhs.ldata_, std::move(rhs.desc_)}, storage_(std::move(rhs.storage_))
   { }
 
+  template <typename O>
+  __MATX_INLINE__ bool isSameView(const O &o) {
+    if constexpr (is_tensor_view_v<O> && RANK == O::Rank()) {
+      return Data() == o.Data() &&
+        this->Shape() == o.Shape() &&
+        this->desc_.Strides() == o.desc_.Strides();
+    } else {
+      return false;
+    }
+  }
   /** Perform a shallow copy of a tensor view
    *
    * Alternative to operator= since it's used for lazy evaluation. This function
@@ -235,7 +245,7 @@ public:
 
   __MATX_INLINE__  ~tensor_t() = default;
 
-  const std::string str() {
+  const std::string str() const {
     return std::string("T") + std::to_string(RANK) + "_" + detail::to_short_str<T>();
   }
 
@@ -896,7 +906,7 @@ public:
     Desc new_desc{this->desc_.Shape(), std::move(strides)};  
     return tensor_t<Type, RANK, Storage, Desc>{storage_, std::move(new_desc), data};
   }
-
+  
   /**
    * Permute the dimensions of a tensor
    *
@@ -912,7 +922,7 @@ public:
    * @returns tensor view of only imaginary-valued components
    *
    */
-  __MATX_INLINE__ auto Permute(const uint32_t (&dims)[RANK]) const
+  __MATX_INLINE__ auto Permute(const std::array<int32_t, RANK> &dims) const
   {
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
     
@@ -937,6 +947,27 @@ public:
     return tensor_t<T, RANK, Storage, Desc>{storage_, std::move(new_desc), this->ldata_};
   }
 
+
+  /**
+   * Permute the dimensions of a tensor
+   *
+   * Accepts any order of permutation. Number of dimensions must match RANK of
+   * tensor
+   *
+   * @tparam M
+   *   Rank of tensor to permute. Should not be used directly
+   *
+   * @param dims
+   *   Dimensions of tensor
+   *
+   * @returns tensor view of only imaginary-valued components
+   *
+   */
+  __MATX_INLINE__ auto Permute(const int32_t (&dims)[RANK]) const
+  {
+    return Permute(detail::to_array(dims));
+  }
+
   /**
    * Permute the last two dimensions of a matrix
    *
@@ -958,7 +989,7 @@ public:
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
     
     static_assert(RANK >= 2, "Only tensors of rank 2 and higher can be permuted.");
-    uint32_t tdims[RANK];
+    int32_t tdims[RANK];
     std::iota(std::begin(tdims), std::end(tdims), 0);
     std::swap(tdims[RANK - 2], tdims[RANK - 1]);
     return Permute(tdims);
--- a/include/matx/core/utils.h
+++ b/include/matx/core/utils.h
@@ -146,5 +146,36 @@ __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ auto madd( const T1 &x, const T2 &
   }
 }
 
+template <int RANK, int D>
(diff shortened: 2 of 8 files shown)


include/matx/core/tensor.h
```diff
@@ -217,6 +217,16 @@ class tensor_t : public detail::tensor_impl_t<T,RANK,Desc> {
       : detail::tensor_impl_t<T, RANK, Desc>{rhs.ldata_, std::move(rhs.desc_)}, storage_(std::move(rhs.storage_))
   { }
 
+  template <typename O>
+  __MATX_INLINE__ bool isSameView(const O &o) {
```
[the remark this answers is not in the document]
luitjens: It's not clear to me if it is needed on the device but maybe.

luitjens:
ready for full review.

include/matx/core/tensor.h
```diff
@@ -896,6 +906,47 @@ class tensor_t : public detail::tensor_impl_t<T,RANK,Desc> {
...
+                      "Cannot list the same dimension to permute twice");
+      done[d] = true;
+      n[i] = this->Size(d);
+      s[i] = this->Stride(d);
+    }
+
+    Desc new_desc{std::move(n), std::move(s)};  
+    return tensor_t<T, RANK, Storage, Desc>{storage_, std::move(new_desc), this->ldata_};
+  }
+
```
cliffburdick, line 949: The C version below should be changed to use this one to avoid code duplication

include/matx/core/utils.h
```diff
@@ -146,5 +146,36 @@ __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ auto madd( const T1 &x, const T2 &
   }
 }
 
+template <int RANK, int D>
+auto __MATX_INLINE__ getPermuteDims( const int (&dims)[D] ) {
```
cliffburdick, line 150: Function case isn't matching the other code. Maybe `GetPermuteDims`?

include/matx/transforms/fft.h
```diff
@@ -608,7 +609,7 @@ class matxFFTPlan2D_t : public matxFFTPlan_t<OutTensorType, InTensorType> {
       auto i_shape = i.Shape();
       // Get total number of batches
       size_t total_iter = std::accumulate(i_shape.begin(), i_shape.begin() + InTensorType::Rank() - batch_offset, 1, std::multiplies<shape_type>());
-      for (size_t iter = 0; iter < total_iter; iter++) {
+      for (size_t iter = 0; iter < total_iter-1; iter++) {
```
cliffburdick, line 612: I think rather than modifying this loop total_iter should be modified above maybe? It should be returning the product of all the dims you want to batch over, so I'm not sure I understand the `-1`.

include/matx/transforms/fft.h
```diff
@@ -748,6 +749,53 @@ auto  GetFFTInputView([[maybe_unused]] OutputTensor &o,
...
+__MATX_INLINE__ auto getCufft2DSupportedTensor( const TensorOp &in, cudaStream_t stream) {
+
+  constexpr int RANK=TensorOp::Rank();
+
+  if constexpr ( !is_tensor_view_v<TensorOp>) {
+    return make_tensor<typename TensorOp::scalar_type>(in.Shape(), MATX_ASYNC_DEVICE_MEMORY, stream); 
+  } else {
+    bool supported = true;
+
+    if ( in.Stride(RANK-2) != in.Stride(RANK-1) * in.Size(RANK-1)) {
```
cliffburdick, line 783: Maybe add a comment here so we can refer back to what is/isn't supported?

include/matx/transforms/fft.h
```diff
@@ -757,55 +805,103 @@ auto  GetFFTInputView([[maybe_unused]] OutputTensor &o,
...
+ *
+ * Creates a new FFT plan in the cache if none exists, and uses that to execute
+ * the 1D FFT. Note that FFTs and IFFTs share the same plans if all dimensions
+ * match
+ *
+ * @tparam OutputTensor
+ *   Output tensor or operatortype
+ * @tparam InputTensor
+ *   Input tensor or operator type
+ * @tparam RANK
```
cliffburdick, line 880: This param is missing now

include/matx/transforms/fft.h
```diff
@@ -815,109 +911,209 @@ void fft(OutputTensor &o, const InputTensor &i,
...
  *
- * @tparam T1
- *   Output view data type
- * @tparam T2
- *   Input view data type
+ * @tparam OutputTensor
+ *   Output operator or tensor
+ * @tparam InputTensor
+ *   Input operator or tensor
  * @tparam RANK
```
cliffburdick, line 1025: RANK can be removed (I know this isn't yours)

cliffburdick (approved):

include/matx/core/tensor.h
luitjens: I tried that but there apparently is a sign change from uint to int and I didn't want to include a fix for that.  I think we probably need to have 2 APIs one for signed and unsigned unfortunately.

include/matx/core/utils.h
luitjens: This is just a movement of what was already in the code.  We should decide on the actual naming convention we should use.  But that might be outside of the scope of this bug.

include/matx/core/tensor.h
luitjens: __MATX_INLINE__ auto Permute(const std::array<int32_t, RANK> &dims) const
 __MATX_INLINE__ auto Permute(const uint32_t (&dims)[RANK]) const

Second is unsigned.  We can change that to signed but then some of our examples/uses were not compiling.  I'm guessing we are hitting some odd c++ rules about literals being signed vs unsigned.

luitjens: so the second above cannot call detail::to_array as it would produce a uint32_t array.

include/matx/transforms/fft.h
luitjens: that -1 is a bug.  I was playing with that to see why we were hitting OOB issues.  I'll fix this.

include/matx/transforms/fft.h
```diff
@@ -203,17 +203,17 @@ template <typename OutTensorType, typename InTensorType> class matxFFTPlan_t {
...
       }
 
       params.batch = (RANK == 2) ? 1 : i.Size(RANK - 3);
       params.inembed[1] = o.Size(RANK-1);
       params.onembed[1] = i.Size(RANK-1);
       params.istride = i.Stride(RANK-1);
       params.ostride = o.Stride(RANK-1);
-      params.idist = i.Size(RANK-2) * i.Size(RANK-1);
-      params.odist = o.Size(RANK-2) * o.Size(RANK-1);
+      params.idist = (RANK<=2) ? 1 : (int) i.Stride(RANK-3);
```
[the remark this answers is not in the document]
luitjens: if RANK==2 and we are doing a 2D FFT then the batch size = 1 and idist is irrelevant.  It should actually be 0 but cufft doesn't support that.  so we put a 1 here.

include/matx/transforms/fft.h
luitjens: fixed

include/matx/transforms/fft.h
luitjens: fixed

include/matx/transforms/fft.h
luitjens: fixed

include/matx/core/tensor.h
luitjens: ok changing to int fixed.  I don't think we need a uint version.

PR #342, opened 2022-12-05
fixing some valgrind warnings

luitjens:


luitjens:
@cliffburdick to check cub fixes.  @tylera-nvidia to check nvtx fixes

tylera-nvidia (approved):

include/matx/transforms/cub.h
```diff
@@ -304,10 +310,15 @@ class matxCubPlan_t {
...
     else if constexpr (op == CUB_OP_INC_SUM || op == CUB_OP_HIST_EVEN) {
       params.batches = TotalSize(a) / a.Size(a.Rank() - 1);
+    } else {
+      params.batches = 1;
     }
+
     params.a_out = a_out.Data();
     params.dtype = TypeToInt<T1>();
 
+    params.stream = stream;
```
cliffburdick, line 320: stream is usually set in the function call itself, so this should be redundant. do you know which one it was complaining about?

luitjens: I deleted the stream sets in the functions and moved it all to this function to avoid redundancy and potential errors.  The valgrind warning I believe was coming from params.size[] not being set when Rank() < 2.

PR #343, opened 2022-12-06
adding more error checking for operator sizes

luitjens:


include/matx/core/error.h
```diff
@@ -225,4 +233,14 @@ namespace matx
     MATX_CUDA_CHECK(e);                \
   }
 
+// This macro asserts compatible dimensions of current class to an operator.
+#define ASSERT_COMPATIBLE_OP_SIZES(op)                               \
+  if constexpr (Rank() > 0) {                                        \
+    _Pragma("unroll")                                                \
```
cliffburdick, line 239: Where is this form of pragma defined?

luitjens: This was defined in c99 and adopted by c++11.  It is a portable way to put a pragma in a macro.

PR #351, opened 2022-12-21
Convolution overhaul.

luitjens:
This version is functionally correct.  Supports all batching options, switching signal and filter, operators, etc.

Performance optimized version will come later.

### DIFF
--- a/include/matx/kernels/conv.cuh
+++ b/include/matx/kernels/conv.cuh
@@ -53,12 +53,13 @@ __global__ void Conv1D(OutType d_out, InType d_in, FilterType d_filter,
   using ftype_strip = typename FilterType::scalar_type;
   using intype_strip = typename InType::scalar_type;
   using outtype_strip = typename OutType::scalar_type;
-  int batch_idx = blockIdx.x;
   uint32_t filter_len = d_filter.Size(Rank-1);
   uint32_t full_len = signal_len + filter_len - 1;
 
+  int batch_idx = blockIdx.x;
+
   // All but the last dim will be populated
-  auto bdims = BlockToIdx(d_in, batch_idx, 1);
+  auto bdims = BlockToIdx(d_out, batch_idx, 1);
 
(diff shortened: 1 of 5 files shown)


include/matx/kernels/conv.cuh
```diff
@@ -187,91 +188,90 @@ __global__ void Conv1D(OutType d_out, InType d_in, FilterType d_filter,
...
-                    tid_x - d_filter.Size(1) + 1 + x), val);
-      }
-      else if constexpr (d_in.Rank() == 2) {
-        val = detail::madd(s_filter[y * d_filter.Size(1) + x],
-               d_in(tid_y - d_filter.Size(0) + 1 + y,
-                    tid_x - d_filter.Size(1) + 1 + x), val);
-      }
-    }
-  }
+  // TODO detect mode based on output size?
```
cliffburdick, line 213: I think it's better to be explicit how you have it and just error if the output size is wrong. it gives us flexibility in the future to do dynamic tensor allocations

include/matx/transforms/conv.h
```diff
@@ -98,53 +98,30 @@ inline void matxDirectConv1DInternal(OutputType &o, const InType &i,
...
 {
-  MATX_STATIC_ASSERT(OutputType::Rank() == InType::Rank(), matxInvalidDim);
-  MATX_STATIC_ASSERT(FilterType::Rank() == 2, matxInvalidDim);
-
-  using filter_input_t = typename FilterType::scalar_type;
-  auto shmsize = sizeof(filter_input_t) * filter.Size(0) * filter.Size(1);
+  MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
+  
+  MATX_STATIC_ASSERT(OutputType::Rank() == In1Type::Rank(), matxInvalidDim);
+  MATX_STATIC_ASSERT(OutputType::Rank() == In2Type::Rank(), matxInvalidDim);
```
cliffburdick, line 109: eventually it would be good to remove this to allow batching a filter without cloning

cliffburdick (approved):

Issue #353, opened 2023-01-04
[FEA] Allow Real * Complex GEMMs


cliffburdick:
Currently BLAS only supports GEMMs with both inputs of the same type (real/complex). We should support a mix of real/complex inputs even if it takes a conversion internally.

luitjens:
Can this be done at the API level with a simple constexpr branch checking types and dispatching the compex operator?


void gemm(...)
if constexp( is_complex_v(Atype) && !is_complex_v(Btype)) {
    gemm(C, A, complex(B), stream);
}  else if constexpr (!is_complex_v(Atype) && is_complex_v(Btype){
    gemm(C, complex(A), B, stream);
} else {
   // existing launch code.
}

cliffburdick:
we need to async allocate a new tensor and upconvert to the required type. For example, if it's complex * real, we need to allocate space for a complex, assign the real to the complex, then do the gemm. I think what you're suggesting would only work if the GEMM libraries could take operators.

luitjens:
I think i wrote it so gemm does this automatically if one of the inputs is an operator.

luitjens:
https://github.com/NVIDIA/MatX/blob/main/include/matx/transforms/matmul.h#L1072-L1087

cliffburdick:
Great, so then we just need to extend that to promote any types to the output type

kshitij12345:
@cliffburdick @luitjens I would like to give it a shot. Will try to send a PR soon :)

PR #362, opened 2023-01-24
matmul: support real/complex tensors

kshitij12345:
Fixes : https://github.com/NVIDIA/MatX/issues/353

### DIFF
--- a/include/matx/transforms/matmul.h
+++ b/include/matx/transforms/matmul.h
@@ -1072,22 +1072,24 @@ void matmul(TensorTypeC C, const TensorTypeA A,
 {
   MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
+  auto A_ = as_type<typename TensorTypeC::scalar_type>(A);
+  auto B_ = as_type<typename TensorTypeC::scalar_type>(B);
   // CublasLt does not support operators and certain transpose modes.
-  // Gab a suppported tensor here and copy in if necessary.
+  // Grab a suppported tensor here and copy in if necessary.
   auto c = getCublasSupportedTensor(C, stream);
-  auto a = getCublasSupportedTensor(A, stream);
-  auto b = getCublasSupportedTensor(B, stream);
+  auto a = getCublasSupportedTensor(A_, stream);
+  auto b = getCublasSupportedTensor(B_, stream);
 
   typedef decltype(c) ctype;
   typedef decltype(a) atype;
   typedef decltype(b) btype;
 
-  if(!a.isSameView(A)) {
-    (a = A).run(stream);
+  if(!a.isSameView(A_)) {
+    (a = A_).run(stream);
   }
 
-  if(!b.isSameView(B)) {
-    (b = B).run(stream);
+  if(!b.isSameView(B_)) {
+    (b = B_).run(stream);
   }
 
 #if MATX_ENABLE_CUTLASS != 1
--- a/include/matx/transforms/matmul.h
+++ b/include/matx/transforms/matmul.h
@@ -1072,6 +1072,13 @@ void matmul(TensorTypeC C, const TensorTypeA A,
 {
   MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
+  constexpr auto is_a_complex = is_complex_v<TensorTypeA>;
+  constexpr auto is_b_complex = is_complex_v<TensorTypeB>;
+  constexpr auto is_c_complex = is_complex_v<TensorTypeC>;
+  
+  // Disallow: A and B != complex and C = complex 
+  static_assert(!((!is_a_complex && !is_b_complex) && is_c_complex));
+
   // promote A and B to the type of C
   auto A_ = as_type<typename TensorTypeC::scalar_type>(A);
   auto B_ = as_type<typename TensorTypeC::scalar_type>(B);
--- a/include/matx/transforms/matmul.h
+++ b/include/matx/transforms/matmul.h
@@ -1072,12 +1072,13 @@ void matmul(TensorTypeC C, const TensorTypeA A,
 {
   MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
-  constexpr auto is_a_complex = is_complex_v<TensorTypeA>;
-  constexpr auto is_b_complex = is_complex_v<TensorTypeB>;
-  constexpr auto is_c_complex = is_complex_v<TensorTypeC>;
-  
-  // Disallow: A and B != complex and C = complex 
-  static_assert(!((!is_a_complex && !is_b_complex) && is_c_complex));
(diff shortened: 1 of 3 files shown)


test/00_transform/MatMul.cu
```diff
@@ -66,8 +66,31 @@ template <typename TensorType>
...
+template <typename TensorType>
+class MatMulTestFloatNonComplexTypes : public MatMulTest<TensorType> {
+};
+
 TYPED_TEST_SUITE(MatMulTestFloatTypes, MatXFloatTypes);
 TYPED_TEST_SUITE(MatMulTestFloatNonHalfTypes, MatXFloatNonHalfTypes);
+TYPED_TEST_SUITE(MatMulTestFloatNonComplexTypes, MatXFloatNonComplexTypes);
+
+template <typename T>
+struct float_to_complex_t
```
kshitij12345, line 78: Is there a similar helper already?

kshitij12345:
cc: @cliffburdick @luitjens

include/matx/transforms/matmul.h
```diff
@@ -1072,22 +1072,26 @@ void matmul(TensorTypeC C, const TensorTypeA A,
 {
   MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
+  // promote A and B to the type of C
+  auto A_ = as_type<typename TensorTypeC::scalar_type>(A);
```
cliffburdick, line 1076: I don't think this will work for half precision mixed with full precision (yet). you could add a static assert to catch that and for now just handle the case where they're both fp32/fp64.

test/00_transform/MatMul.cu
cliffburdick: I don't think it exists yet. Thanks!

test/00_transform/MatMul.cu
```diff
@@ -66,8 +66,31 @@ template <typename TensorType>
...
+template <>
+struct float_to_complex_t<matxFp16>
+{
+  using t = matxFp16Complex;
+};
+
+template <>
+struct float_to_complex_t<matxBf16>
+{
+  using t = matxBf16Complex;
```
cliffburdick, line 92: Maybe call these `float_to_complex` and add a helper function that does `using float_to_complex_t = ...` to alias it and remove the need for ::t. Also can you use `type` instead of `t`?

test/00_transform/MatMul.cu
```diff
@@ -325,6 +348,58 @@ TYPED_TEST(MatMulTestFloatTypes, MediumRectBatched3DStridedBatch)
   MATX_EXIT_HANDLER();
 }
 
+TYPED_TEST(MatMulTestFloatNonComplexTypes, MixedTypes)
+{
+  using ComplexTypeParam = typename float_to_complex_t<TypeParam>::t;
+  // a -> real, b -> real, c -> complex
```
cliffburdick, line 354: a * b resulting in a complex number should not be valid, should it?

test/00_transform/MatMul.cu
```diff
@@ -325,6 +348,58 @@ TYPED_TEST(MatMulTestFloatTypes, MediumRectBatched3DStridedBatch)
...
+
+    constexpr index_t m = 4;
+    constexpr index_t k = 8;
+    constexpr index_t n = 16;
+
+    tensor_t<TypeParam, 2> a{{m, k}};
+    tensor_t<ComplexTypeParam, 2> b{{k, n}};
+    tensor_t<ComplexTypeParam, 2> c{{m, n}};
+
+    this->pb->template InitAndRunTVGenerator<TypeParam>(
```
cliffburdick, line 390: Without modifying the Python script, I'm not sure how this is working. This is using `TypeParam` (real) as the data type passed to Python. Inside of Python I think this is just doing a real * real, resulting in a real, but you are treating a real tensor as a complex. I think `b` has no imaginary values. Can you confirm?

cliffburdick:
> cc: @cliffburdick @luitjens

Thanks @kshitij12345 for doing all the work! It looks good, but I left a few comments to help clarify things for me.

test/00_transform/MatMul.cu
kshitij12345: Interestingly numpy allows that. (Though I agree that it feels weird).
```python
import numpy as np

a = np.ones((3, 4))
b = np.ones((4, 3))
c = np.ones((3, 3), dtype=np.complex128)

np.matmul(a, b, out=c)
print(c)
```

Do you think we should restrict this case?

kshitij12345:
@cliffburdick Thanks for the review. Will most likely address it by or before this weekend :)

test/00_transform/MatMul.cu
cliffburdick: I'm not entirely sure what the use case for that would be other than you wanting to use `c` as a complex later on, but it causes more overhead to convert both `a` and `b` to complex during the GEMM, rather than just converting `c` to complex when it's needed. I don't have a very strong opinion either way, but it doesn't seem like something we'd want to encourage people to do. I think the usefulness for this feature is mixing real/complex a/b since that's used extremely often in signal processing.

cliffburdick:
> @cliffburdick Thanks for the review. Will most likely address it by or before this weekend :)

Thanks!

test/00_transform/MatMul.cu
kshitij12345: Yes. You are right. Would there be a simple way to check it? From what I see and understand `InitAndRunTVGenerator` always produces all np.arrays corresponding to TypeParam.

Thanks for catching this!

test/00_transform/MatMul.cu
kshitij12345: Makes sense! Also, `restricting first and then allowing` would be easier than `allowing and then having to restrict it` (for whatever reason).

Will disallow this case. Thanks!

test/00_transform/MatMul.cu
kshitij12345: Updated.

include/matx/transforms/matmul.h
kshitij12345: Sorry but could you elaborate which mixed precision won't work? I am not sure about the same. Thanks

kshitij12345:
Have updated for the PR based on all the comments barring one. PTAL. Thank you!

include/matx/transforms/matmul.h
cliffburdick: Maybe it will now that I rethink that logic. It would have to be tested, but I'll close the comment for now.

cliffburdick (approved):

kshitij12345:
Gentle ping :)

cliffburdick:
> Gentle ping :)

hi @kshitij12345 , I will test this out first thing tomorrow and merge if all is well

kshitij12345:
Sure. Thank you :)

PR #363, opened 2023-01-26
Slice support for std::array shapes

luitjens:


include/matx/operators/slice.h
```diff
@@ -250,14 +285,24 @@ namespace matx
    * @param ends the last element (exclusive) of each dimension of the input operator.  matxDrop Dim removes that dimension.  matxEnd deontes all remaining elements in that dimension.
    * @return sliced operator
    */
+  template <int N, typename OpType>
+  __MATX_INLINE__ auto slice (const OpType opIn, 
+      const std::array<index_t, OpType::Rank()> &starts,
+      const std::array<index_t, OpType::Rank()> &ends)
+  {
+    std::array<index_t, OpType::Rank()> strides;
+    for(int i = 0; i < OpType::Rank(); i++)
```
cliffburdick, line 294: `strides.fill`?

cliffburdick: Although. std::array isn't officially in libcudacxx until later versions.

include/matx/operators/slice.h
```diff
@@ -250,14 +285,24 @@ namespace matx
...
+      strides[i] = 1;
```
cliffburdick, line 295: I think I'm missing what setting this to 1 does. I don't see it recomputing the strides in StrideOp, so where is it setting the correct strides?

luitjens: These are strides to the operator which doesn't have the concept of strides like the tensor does.  Though it seems like there could be an issue with our stride optimization applying to the tensor directly.  I might have to duplicate a little code to handle that.  IIUC no stride in operators = all ones.  no stride change in tensors = -1 in first spot.  Agree?

include/matx/operators/slice.h
luitjens: This didn't seem to impact compile or correctness.  I think these are on host only thus not an issue?

cliffburdick: Yeah, so can it be switched to fill to save a line?

include/matx/operators/slice.h
cliffburdick: I don't see the strides being set correctly anywhere, but I see this usage: 
```
            for(int i = 0; i < Rank(); i++) {
              ind[dims_[i]] += inds[i] * strides_[i]; 
            }
```

So if you pass in all 1, won't this be wrong unless the strides are computed correctly like we do on tensors?

tylera-nvidia (approved):

cliffburdick (approved):

include/matx/operators/slice.h
luitjens: no strides being set to 1 means no changes in the logical strides.  In the tensor class this just sets stride_mult=1 which means the underlying storage strides don't get modified.  operators have no underlying strides.  The stride get's applied as just multiplying the incoming indices.  so stride=2:  index 0,1,2,3 -> 0,2,4,6

include/matx/operators/slice.h
luitjens: fixed

PR #369, opened 2023-02-02
Adding 2 new operators:

luitjens:
1)  sign(x)  this returns -1 if negative, 0 if 0, and 1 if positive. 2)  index(dim) this returns the index of the tensor in the provided dim.

2) required some infastructure changes to support a rankless and sizeless operator.  I added a sentinal maxNoRank to signal this.

### DIFF
--- a/include/matx/core/type_utils.h
+++ b/include/matx/core/type_utils.h
@@ -52,6 +52,10 @@
 
 namespace matx {
 
+enum {
+  matxNoRank = -1
+};
+
 enum class MemoryLayout {
   MEMORY_LAYOUT_ROW_MAJOR,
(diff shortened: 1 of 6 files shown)


include/matx/core/type_utils.h
```diff
@@ -52,6 +52,10 @@
 
 namespace matx {
 
+enum {
+  matxNoRank = -1
```
cliffburdick, line 56: shouldn't this be defined with the rest of the things like `matxEnd`?

luitjens: I don't think so.  matxEnd is a different concept entirely used for clone.  this has nothing to do with clone.

cliffburdick (approved):

PR #373, opened 2023-02-07
implement QR for small batched matrices.

luitjens:
Rename cusparse wrapper to cusparse_qr.
Cusolver API doesn't currently produce QR but instead R and the HH vectors.

### DIFF
--- a/include/matx/transforms/qr.h
+++ b/include/matx/transforms/qr.h
@@ -1,0 +1,223 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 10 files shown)


include/matx/transforms/qr.h
```diff
@@ -0,0 +1,248 @@
...
+             v = x)).run(stream);
+    
+    //printf("v:\n"); Print(v);
+
+    sum(ci, norm(v), stream);
+    //printf("s:\n"); Print(ci);
+
+    (ci = NTypeS(2) / ci).run(stream);
+
+    //printf("ci:\n"); Print(ci);
```
cliffburdick, line 206: remove prints

test/00_solver/QR2.cu
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  cudaDeviceSynchronize();
+
+  ASSERT_NEAR( mdiffQTQ(), SType(0), .00001);
+  ASSERT_NEAR( mdiffQR(), SType(0), .00001);
+}
+
+TYPED_TEST(QR2SolverTestNonHalfTypes, QR2)
+{
```
cliffburdick, line 115: we should eventually move to INSTANTIATE_TEST_CASE_P. this will allow real parameter sweeps that have correct errors printed

cliffburdick (approved):

include/matx/transforms/qr.h
luitjens: fixed

Issue #365, opened 2023-01-27
[BUG] CUB does not support strided tensors.


luitjens:
Cub does not appear to support strided tensors (at least on output).  We should either add support for that,  assert the incompatibility, or  find another work around.

PR #380, opened 2023-02-10
Added output iterator support for CUB sums, and converted all sum()

### LINKED ISSUE #311: [FEA] support operators for output in cub_reduce
Cub reduce should support operators on output.  It currently calls .Data().

cliffburdick:
closes #311  #365


CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ include(cmake/FindLibcudacxx.cmake)
 target_include_directories(matx INTERFACE "$<BUILD_INTERFACE:${LIBCUDACXX_INCLUDE_DIR}>")    
 
 # Set flags for compiling tests faster
-set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --threads 0)
+set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --threads 0 -ftemplate-backtrace-limit 100)
```
luitjens, line 107: why not set it to 0 which gives unlimited?

include/matx/core/tensor.h
```diff
@@ -73,7 +73,7 @@ enum {
 #endif
 
 /**
- * @brief Iterator for TensorImpl for libraries that can take iterators as input (CUB).
+ * @brief Iterator around operators for libraries that can take iterators as input (CUB).
```
luitjens, line 76: should we move RandomOperatorIterator to its own source file?  It doesn't seem like it belongs in tensor.h

cliffburdick: Good idea. I was also going to combine the two iterators since we don't need two. One that has both l/rvalue operator() defined would suffice

CMakeLists.txt
luitjens: @cliffburdick should we set this to 0?  or do you think 100 is enough?

include/matx/core/iterator.h
```diff
@@ -0,0 +1,230 @@
...
+  {
+    return self_type{t_, offset_ + offset};
+  }
+
+  [[nodiscard]] __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ value_type operator[](difference_type offset) const
+  {
+    return *self_type{t_, offset_ + offset};
+  }  
+
+  __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__  self_type operator++(int)
```
luitjens, line 91: is this correct?  wouldn't ++(int) increment by int?

include/matx/core/iterator.h
```diff
@@ -0,0 +1,230 @@
...
+/**
+ * @brief Iterator around operators for libraries that can take iterators as output (CUB).
+ * 
+ * @tparam T Data type
+ * @tparam RANK Rank of tensor
+ * @tparam Desc Descriptor for tensor
+ * 
+ */
+template <typename OperatorType>
+struct RandomOperatorOutputIterator {
```
luitjens, line 146: Do we need both input and output operator or can we have one with both lvalue and rvalue semantics?

include/matx/core/iterator.h
```diff
@@ -0,0 +1,230 @@
...
+namespace matx {
+/**
+ * @brief Iterator around operators for libraries that can take iterators as input (CUB).
+ * 
+ * @tparam T Data type
+ * @tparam RANK Rank of tensor
+ * @tparam Desc Descriptor for tensor
+ * 
+ */
+template <typename OperatorType>
```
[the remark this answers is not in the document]
cliffburdick: I thought about this but ended up leaving it out. The reason is I could foresee scenarios where a user would want to use this iterator type directly in their own code. I'm fine wrapping it if you don't think that'll happen

include/matx/core/iterator.h
cliffburdick: Not sure what you mean, but this is incrementing by integer on offset_, but it returns the old value (iterator)

include/matx/core/iterator.h
cliffburdick: I tried getting rid of the output version and got a ton of compiler warnings. I'll leave it as TBD for now

include/matx/core/iterator.h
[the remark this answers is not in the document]
luitjens: i guess I look at detail as internal API that is subject to change and matx as namespace we want to get stable.  I'm fine with either.  I'm just not sure a user would actually use this API but maybe.

include/matx/core/iterator.h
luitjens: oh right i had forgot what operator++(int) was.  it's pre vs post ++.

luitjens (approved):

PR #388, opened 2023-03-09
Fixes for 32b mode

cliffburdick:
Fixes for 32b mode and disable CUB caching


include/matx/transforms/fft.h
```diff
@@ -824,7 +824,7 @@ __MATX_INLINE__ auto getCufft2DSupportedTensor( const TensorOp &in, cudaStream_t
  */
 template <typename OutputTensor, typename InputTensor>
 __MATX_INLINE__ void fft(OutputTensor o, const InputTensor i,
-         index_t fft_size = 0, cudaStream_t stream = 0)
+         uint64_t fft_size = 0, cudaStream_t stream = 0)
```
luitjens, line 827: do we want a comment on why uint64_t is necessary instead of index_t?

luitjens (approved):

PR #400, opened 2023-03-30
Allow 0-stride dimensions for cublas input/output

tbensonatl:
Currently, if a tensor contains 0-stride dimensions in one of the last two dimensions, then that tensor will not be used directly as input/output to cublas matmul and a copy will be made transparently. This generates copies during matvec operations because matvec clones a tensor and delegates to matmul, which then copies the tensor because the clone creates a 0-stride dimension. This change allows 0-stride dimensions if the dimension size is 1 (and thus no index will stride in that dimension) and thus eliminates allocations/frees and copies when calling matvec.

### DIFF
--- a/test/00_transform/MatMul.cu
+++ b/test/00_transform/MatMul.cu
@@ -561,3 +561,107 @@ TYPED_TEST(MatMulTestFloatNonHalfTypes,  MatMulOp)
 
   MATX_EXIT_HANDLER();
 }
+
+TYPED_TEST(MatMulTestFloatTypes, MediumMatVec)
+{
+  MATX_ENTER_HANDLER();
+  constexpr index_t m = 128;
+  constexpr index_t k = 256;
+  constexpr index_t n = 1;
+
+  tensor_t<TypeParam, 2> a{{m, k}};
+  tensor_t<TypeParam, 2> b{{k, n}};
+  tensor_t<TypeParam, 2> c{{m, n}};
+  this->pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "matmul_operators", "run", {m, k, n});
+
+  this->pb->NumpyToTensorView(a, "a");
+  this->pb->NumpyToTensorView(b, "b");
+
+  auto cs = c.template Slice<1>({0,0}, {matxEnd, matxDropDim});
+  auto bs = b.template Slice<1>({0,0}, {matxEnd, matxDropDim});
+  matvec<decltype(cs), decltype(a), decltype(bs), PROVIDER_TYPE_CUBLASLT>(cs, a, bs);
+
+  MATX_TEST_ASSERT_COMPARE(this->pb, c, "c", this->thresh);
+
+  // Test also with rank-1 tensors rather than just slices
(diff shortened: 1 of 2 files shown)


test/00_transform/MatMul.cu
```diff
@@ -561,3 +561,107 @@ TYPED_TEST(MatMulTestFloatNonHalfTypes,  MatMulOp)
...
+  tensor_t<TypeParam, 2> a{{m, k}};
+  tensor_t<TypeParam, 2> b{{k, n}};
+  tensor_t<TypeParam, 2> c{{m, n}};
+  this->pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "matmul_operators", "run", {m, k, n});
+
+  this->pb->NumpyToTensorView(a, "a");
+  this->pb->NumpyToTensorView(b, "b");
+
+  auto cs = c.template Slice<1>({0,0}, {matxEnd, matxDropDim});
```
cliffburdick, line 581: you can use `slice` here instead since it's the new preferred method and gets rid of the need for the template keyword

cliffburdick:
thanks for adding all the tests!

test/00_transform/MatMul.cu
tbensonatl: @cliffburdick Using `auto cs = slice({0,0}, {matxEnd, matxDropDim});`? Will that result in a rank 2 tensor `cs` with `Size(1) == 0`? In this case, the subsequent `matvec` requires that `b` and `c` both have rank one smaller than `a`. Or perhaps we can generalize `slice` to rank-reduce when using `matxDropDim`.

cliffburdick: the first parameter to `slice` is the input operator. other than that, the arguments are the same. it should result in exactly the same thing you had, but the syntax is nicer and we're trying to move away from the member functions

tbensonatl: Oh, sorry, I meant `auto cs = slice(c, {0,0}, {matxEnd, matxDropDim});`. But I think that will give use the following:

tylera-nvidia: I believe you'll need to use the [matx::slice with Rank as a template param](https://github.com/NVIDIA/MatX/blob/522c7419648f4a4ca339a6234ba5adda52393882/include/matx/operators/slice.h#L247), as the slice with no templates params will always return with the same rank:

```
    auto c = matx::make_tensor<float,2>({16,16});
    auto cs = matx::slice<1>(c, {0,0}, {matx::matxEnd, matx::matxDropDim});
    std::cout << "Rank of CS: " << cs.Rank() << std::endl;  
```
 output:
 
```
 docker@e5c559bbda66:/home/tallen/scratch/playground/matxExample/build$ ./matxExample 
Rank of CS: 1
```

tylera-nvidia (approved):

test/00_transform/MatMul.cu
tbensonatl: @tylera-nvidia Thanks, I missed that `slice` still had a rank template parameter. I have updated the tests to use `slice` instead of `Slice`.

PR #406, opened 2023-04-13
Add Mutex to support multithread NVTX markers

tylera-nvidia:


include/matx/core/nvtx.h
```diff
@@ -83,8 +85,9 @@ static const int32_t nvtxColors[nunNvtxColors] = {
                                              NVTX_WHITE
                                              };
 
-inline uint64_t curColorIdx;
-inline std::map< int, nvtxRangeId_t>  eventMap;
+inline uint64_t                      curColorIdx;     ///< counter for rotation of colors for sequential ranges
+inline std::map< int, nvtxRangeId_t> nvtx_eventMap;   ///< map of currently active NVTX ranges
+inline std::shared_mutex             nvtx_memory_mtx; ///< Mutex protecting updates from map
```
cliffburdick, line 90: is there a reason this is a `shared_mutex` instead of just a `mutex`?

tylera-nvidia: I noticed we use `shared_mutex` [here](https://github.com/NVIDIA/MatX/blob/4268f090efa950d9ea781cb668e0213d0fd9c2b6/include/matx/core/allocator.h#L83). To your point, I can't imagine a pattern where I would want several threads to be operating on it at once... I can change this one to a standard `mutex`. Should I change the allocator one as well?

cliffburdick: Sorry, not sure why I didn't see this. It seems like a standard mutex would be better since we don't need any type of sharing.

cliffburdick (approved):

PR #407, opened 2023-04-14
Add single argument copy() and copy() tests

tbensonatl:
Add a single argument version of copy() that allocates a new tensor of the same type/size as the tensor being copied and performs a deep copy into the new tensor.

### DIFF
--- a/include/matx/transforms/copy.h
+++ b/include/matx/transforms/copy.h
@@ -60,7 +60,7 @@ namespace matx
     __MATX_INLINE__ void copy(OutputTensor &out, const InputTensor &in,
         Executor exec)
     {
-      MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)      
+      MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
       for (int i = 0; i < OutputTensor::Rank(); i++)
       {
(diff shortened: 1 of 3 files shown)


include/matx/transforms/copy.h
```diff
@@ -69,11 +69,39 @@ namespace matx
...
+   *   Tensor to copy from
+   * @param exec
+   *   Executor to run copy in
+   * @return New tensor into which the input view has been copied
+   */
+  template <typename Tensor, typename Executor, std::enable_if_t<is_executor_t<Executor>(), bool> = true>
+  __MATX_INLINE__ Tensor copy(const Tensor &in, Executor exec)
+  {
+    MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
+    Tensor out({in.Shape()});
```
cliffburdick, line 96: What are the braces here for?

cliffburdick (approved):

include/matx/transforms/copy.h
tbensonatl: Good question - I had gotten a build error early on without them and meant to return to that to investigate. It may have been something wrong in the test cases though, so I'll check again.

tbensonatl: I pushed an update to use Descriptor() rather than Shape()

PR #408, opened 2023-04-17
Add rank0 tensor remap support

tbensonatl:
Add remap() support for 0-rank index tensors

This update treats 0-rank index tensors as if they were rank-1 index tensors of size 1.


luitjens (approved):

luitjens:
@tbensonatl please rebase and squash

cliffburdick (approved):

PR #409, opened 2023-04-19
Fix a few issues highlighted by linters/clang

tbensonatl:
Attempting to build some code with newer versions of Clang highlighted a few issues, including a throw from a noexcept method, an assertion removing a function call in release builds, and some runtime rank-checks used in constexpr expressions.

### DIFF
--- a/include/matx/executors/device.h
+++ b/include/matx/executors/device.h
@@ -84,41 +84,41 @@ namespace matx
           MATX_STATIC_ASSERT((sizeof(op) + sizeof(index_t) * Op::Rank()) <= CUDA_MAX_VAL_PARAM, 
               "Parameter buffer to device is limited to 4096B. Please break up your operator statement into multiple executions to limit the size of the parameters");
 
-          if constexpr (op.Rank() == 0) {
+          if constexpr (Op::Rank() == 0) {
             threads = 1;
             blocks = 1;
             detail::matxOpT0Kernel<<<blocks, threads, 0, stream_>>>(op);
           }
           else {
-            std::array<index_t, op.Rank()> sizes;
-            for (int i = 0; i < op.Rank(); i++) {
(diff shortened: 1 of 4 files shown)


cliffburdick:
@tbensonatl  This is showing a unit test failure:

```
[ RUN      ] OperatorTestsNumericAllExecs/8.SliceOp
/home/jenkins/workspace/unit-tests/test/00_operators/OperatorTests.cu:656: Failure
Expected equality of these values:
  t4t(i, j, k, l)
    Which is: 6
  t4(i + 1, j + 2, k + 3, l + 4)
    Which is: 6
[  FAILED  ] OperatorTestsNumericAllExecs/8.SliceOp, where TypeParam = std::tuple<unsigned long, matx::cudaExecutor> (1 ms)
```

tbensonatl:
@cliffburdick I suspect that the unit test failure was due to a missing synchronization in that test case. I pushed an update to add a synchronization and will restart a CI build and test. I was unable to replicate the failure locally, but I assume that is because it is timing sensitive.

include/matx/executors/device.h
```diff
@@ -84,41 +84,41 @@ namespace matx
           MATX_STATIC_ASSERT((sizeof(op) + sizeof(index_t) * Op::Rank()) <= CUDA_MAX_VAL_PARAM, 
               "Parameter buffer to device is limited to 4096B. Please break up your operator statement into multiple executions to limit the size of the parameters");
 
-          if constexpr (op.Rank() == 0) {
+          if constexpr (Op::Rank() == 0) {
```
cliffburdick, line 87: These are allowed by the standard, but MSVC also complains about it. I think it's a losing proposition to try to fix all of them since there are hundreds in the code.

cliffburdick (approved):

PR #437, opened 2023-06-28
Adding initial polyphase resampler transform

tbensonatl:
Initial version of a polyphase resampler. This version is focused primarily on correctness and a simple interface (i.e. no temporary work buffers).

### DIFF
--- a/include/matx/kernels/resample_poly.cuh
+++ b/include/matx/kernels/resample_poly.cuh
@@ -1,0 +1,211 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its
+//    contributors may be used to endorse or promote products derived from
+//    this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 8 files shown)


include/matx/kernels/resample_poly.cuh
```diff
@@ -0,0 +1,209 @@
...
+#include "matx/core/type_utils.h"
+#include "matx/core/tensor_utils.h"
+
+namespace matx {
+
+#ifdef __CUDACC__ 
+template <int THREADS, typename OutType, typename InType, typename FilterType>
+__launch_bounds__(THREADS)
+__global__ void ResamplePoly1D(OutType output, InType input, FilterType filter,
+                       index_t output_len, index_t input_len,
```
cliffburdick, line 54: Why aren't the lengths known by the operator sizes?

tbensonatl: Good point - I'll remove those parameters.

awthomp:
Thrilled to see this PR! @tbensonatl -- Are you using the critically sampled channelizer here? As a sanity check, it'd be good to confirm correctness between this and our implementation with [cusignal](https://docs.rapids.ai/api/cusignal/stable/api/#module-cusignal.filtering.channelize_poly), if you haven't already done so. I'm curious about speedups too :).

include/matx/kernels/resample_poly.cuh
```diff
@@ -0,0 +1,209 @@
...
+    // need to determine which filter samples we need for this phase and load them
+    // into shared memory.
+    // left_filter_ind is the filter index that corresponds to the input sample
+    // in the upsampled sequence equal to or to the left of the output sample.
+    // Thus, for phase 0, left_filter_ind is the central filter tap, h[k/2].
+    // For phase 1, with our up=3, down=2 example, left_filter_ind is
+    // h[K/2+2]. Once we have a single tap index and the corresponding x index,
+    // we will stride in both sequences (input and filter) by +/- up to cover
+    // the full filter phase.
+    const int left_filter_ind = filter_len_half + (phase_ind*down) % up;
```
cliffburdick, line 114: Does this need to be `index_t`?

include/matx/kernels/resample_poly.cuh
```diff
@@ -0,0 +1,209 @@
...
+
+    // If left_filter_ind >= filter_len, then the filter is not long enough to reach
+    // a potentially non-zero sample value on the left. In that case, set the
+    // last_filter_ind to the next non-zero sample to the right of the output
+    // index.    
+    const int last_filter_ind = (left_filter_ind < filter_len) ?
```
cliffburdick, line 120: `index_t`? a few more places below too

test/00_transform/ResamplePoly.cu
```diff
@@ -0,0 +1,433 @@
...
+  std::unique_ptr<detail::MatXPybind> pb;
+  double thresh;;
+};
+
+template <typename TensorType>
+class ResamplePolyTestFloatTypes
+    : public ResamplePolyTest<TensorType> {
+};
+
+TYPED_TEST_SUITE(ResamplePolyTestFloatTypes, MatXFloatNonHalfTypes);
```
cliffburdick, line 71: Is there a reason half doesn't work? Could it be tested with a large precision?

cliffburdick (approved):
Nice work! Can you add documentation in docs_input as well? At a minimum it can be added to the API section, but if you want to make a new rst file with some text about usage, limitations, etc, that would be useful.

include/matx/kernels/resample_poly.cuh
tbensonatl: I can switch to all index_t in the kernel for now. It will lead to some 64-bit integer divisions in the case of 64-bit index_t, but future optimizations can determine when it can safely be downgraded to 32-bit (e.g. for smem indices).

tbensonatl:
> Thrilled to see this PR, @tbensonatl. I originally posted about the polyphase channelizer, asking if it matched the cusignal version, before realizing this PR is about the _resampler_. My bad, but I'm still equally happy to see MatX include these core signal processing functions.

@awthomp Right, this is a polyphase resampler, but there is also interest in polyphase filters and channelizers. I assume you are talking about this cusignal function: https://docs.rapids.ai/api/cusignal/stable/api/#module-cusignal.filtering.channelize_poly ? I will take a look at that in the future.

include/matx/kernels/resample_poly.cuh
cliffburdick: You can compile with `-DMATX_BUILD_32_BIT` it will change `index_t` to a signed 32b int. We have one project internally that relies on this working. You can also specify custom sizes, but that's not tested very well.

include/matx/kernels/resample_poly.cuh
tbensonatl: Fixed in new commit

test/00_transform/ResamplePoly.cu
tbensonatl: I added some test cases with short filters and enabled half types for those.

PR #452, opened 2023-07-19
Transforms as operators

### LINKED ISSUE #450: [FEA] Investigate mixing transforms and operators
Right now to do an FFT convolution at least three separate lines must be written:
1) FFT input
2) element-wise multiply the output
3) IFFT the multiplied results

This limitation is due to how we treat transforms and operators differently. This issue is to investigate mixing the two, which has many benefits, including saving temporary variables, more readable code, and more chances for optimizations in the parsing.

cliffburdick:
This PR marks a major **breaking change** in MatX syntax. Previously we had a clear distinction of operators vs transforms such that they could not be combined. To do an FFT convolution, the user would write:

```
  // Perform the FFT in-place on both signal and filter
  fft(sig_freq, sig_time);
  fft(filt_freq, filt_time);

  // Perform the pointwise multiply. Overwrite signal buffer with result
  (sig_freq = sig_freq * filt_freq).run();

  // IFFT in-place
  ifft(sig_freq, sig_freq);
```

This code is taken directly from the FFT convolution sample. While the code is readable, anyone who has done an FFT convolution in another language like Python would usually write it as a single line. This was not possible in MatX due to transforms having the requirement of being on their own line. The other thing to notice is that transforms have a different syntax than regular operator expressions; in one case the output is a parameter to a function, while in the other case the output is done using the assignment operator. Having the output as a function parameter has a couple of benefits, such as deducing what to do based on the output, but it has far more downsides.

This PR introduces "transforms as operators". The general idea is that if the user wants to, they can use transforms inside of *any* operator expression. The FFT convolution above changes to: 

```
(sig_freq = ifft(fft(sig_time) * fft(filt_time))).run(stream);
```

Someone familiar with signal processing will immediately recognize the convolution. Besides readability, this syntax has many benefits:

- Fewer temporary variables need to be instantiated for interim steps
- Better operator fusion is possible since more about the entire expression is known at compile-time
- The syntax matches higher-level languages (Python & MATLAB)
- The executor is decoupled from the transform
- The API documentation for each function is cleaner
- Improved NVTX nsys timelines

Under the hood, the feature works by detecting whether a transform is being used as part of a larger expression, and, if so, a temporary async-allocation is made scoped to the lifetime of the operator. In our testing we've found the overhead from these allocations are small (~1%), depending on the size of the problem. Users not wanting the extra overhead can continue to use the old multi-line method, but converted to the new assignment syntax:

```
  (sig_freq = fft(sig_time)).run();
  (filt_freq = fft(filt_time)).run();

  // Perform the pointwise multiply. Overwrite signal buffer with result
  (sig_freq = sig_freq * filt_freq).run();

  // IFFT in-place
  (sig_freq = ifft(sig_freq)).run();
```

In this case MatX detects that these are "simple" transforms, and reverts to the fast-path without any asynchronous allocations, but maintaining the newer syntax.

Transforms that return multiple types are required to be standalone, as they are in Python:

```
(mtie(Evv, Wov) = eig(Bv)).run();
```

`mtie` is used to "tie" multiple values together similar to `std::tie()`. `mtie` is needed to overload the assignment operator.

Feedback is always welcome!

Closes #450


examples/fft_conv.cu
```diff
@@ -104,17 +112,39 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
...
-  filt_time.PrefetchDevice(0);
+  filt_time.PrefetchDevice(0);  
+
 
   // Perform the FFT in-place on both signal and filter
-  fft(sig_freq, sig_time);
-  fft(filt_freq, filt_time);
+  for (int i = 0; i < iterations; i++) {
+    if (i == 1) {
+      cudaEventRecord(start, stream);
```
luitjens, line 121: Is this meant to be a benchmark?  if so maybe put it in bench instead and just leave the example as an example without iterations/timing?

examples/mvdr_beamformer.h
```diff
@@ -116,14 +116,14 @@ class MVDRBeamformer {
     (covMatView = (covMatView * (1.0f / static_cast<float>(snap_len_))) +
                    eye<complex>() * load_coeff_)
         .run(stream);
-    inv(invCovMatView, covMatView, stream);
+    inv_impl(invCovMatView, covMatView, stream);
```
luitjens, line 119: Do we need to rename with _impl?  Or is an overload preferable?

examples/spectrogram_graph.cu
```diff
@@ -55,116 +55,160 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
...
-#endif            
-    }
-  }
-
-  auto f = make_tensor<float>({2000,2000});
-  auto f2 = make_tensor<float>({2000,2000});
-
-  (f = random<float>({2000,2000}, NORMAL)).run(stream);
-  inv(f2, f, stream);
+//   cudaEvent_t start, stop;
```
luitjens, line 64: Why is all those commented code left?  If it is stale let's delete it.

include/matx/core/tie.h
```diff
@@ -0,0 +1,112 @@
...
+  using matxoplvalue = bool;
+
+  static_assert(sizeof...(Ts) > 1, "Only use mtie() when number of elements is greater than 1");
+
+  __MATX_INLINE__ const std::string str() const {
+    if constexpr (sizeof...(Ts) == 2) {
+      return get_type_str(std::get<0>(ts_)) + "=" + get_type_str(std::get<1>(ts_));
+    }
+    else {
+      // Figure this out for proper LHS
```
luitjens, line 58: Is there more work here?  We we can't figure this out now can we add a static assert here to show that this is not working?

include/matx/core/tie.h
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  template <int n>
+  auto get() {
+    return std::get<n>(ts_);
+  }
+
+  static __MATX_INLINE__ constexpr int32_t Rank()
+  {
+    return 1;
```
luitjens, line 87: Why is rank hard coded to 1?

include/matx/core/tie.h
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  constexpr __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto Size([[maybe_unused]] int dim) const noexcept
+  {
+    return 0;
```
luitjens, line 92: Why is size hard coded to 0?

include/matx/operators/dct.h
```diff
@@ -102,7 +102,7 @@ void dct(OutputTensor &out, const InputTensor &in,
 
   tensor_t<cuda::std::complex<typename OutputTensor::scalar_type>, 1> tmp{{N + 1}};
 
-  fft(tmp, in, 0, stream);
+  fft_impl(tmp, in, 0, stream);
```
luitjens, line 105: do we need to rename or can we just overload?

include/matx/transforms/inverse.h
```diff
@@ -377,4 +378,72 @@ void inv(TensorTypeAInv &a_inv, const TensorTypeA &a,
   }
 }
 
+
+namespace detail {
```
luitjens, line 382: Move to operators?

luitjens (changes requested):

examples/mvdr_beamformer.h
cliffburdick: Mentioned on slack, but I prefer _impl because it's easier for docs and explicitly separates operator from transform

examples/spectrogram_graph.cu
cliffburdick: I'm going to revert this file almost entirely once done

include/matx/core/tie.h
cliffburdick: I'll figure it out, but I was getting a compiler issue I wasn't sure if it was a bug. Basically for cases like `((U, S, V) = svd(A))` we want the LHS to print `(TF32, TF32, TF32)` to make it clear there are multiple return types

include/matx/core/tie.h
cliffburdick: I can make it noRank. It doesn't matter since it's never used.

include/matx/core/tie.h
cliffburdick: It's never used.

include/matx/operators/dct.h
cliffburdick: We talked about this on slack, but renaming is intended to be clear what's an implementation detail vs user API

PR #456, opened 2023-07-25
Adds normalization options for `fft` and `ifft`

nvjonwong:
Add normalization scale factors that numpy supports (https://numpy.org/doc/stable/reference/routines.fft.html#normalization). `BACKWARD` is the default, while `ORTHO` and `FORWARD` are as described in the docs above.

- Tests included for non-complex half types.
- Most complex half types work except one particular case.


include/matx/transforms/fft.h
```diff
@@ -956,6 +971,15 @@ __MATX_INLINE__ void ifft(OutputTensor o, const InputTensor i,
   // adjusts size of tensor based on fft_size
   auto in = detail::GetFFTInputView(out, in_t, fft_size, stream);
 
+  // Normalize input if necessary
+  using input_type = decltype(in_t);
+  using s_type = typename detail::value_promote_t<typename input_type::scalar_type>;
+  if (norm == ORTHO) {
+    (in *= std::sqrt(static_cast<s_type>(fft_size))).run(stream);
```
cliffburdick, line 978: The comment says 1/sqrt(N), but this seems to be sqrt(N). which is right?

test/00_transform/FFT.cu
```diff
@@ -92,6 +92,50 @@ TYPED_TEST(FFTTestComplexTypes, FFT1D1024C2C)
...
+  MATX_ENTER_HANDLER();
+  const index_t fft_dim = 1024;
+  this->pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "fft_operators", "fft_1d_fwd", {fft_dim, fft_dim});
+
+  tensor_t<TypeParam, 1> av{{fft_dim}};
+  tensor_t<TypeParam, 1> avo{{fft_dim}};
+  this->pb->NumpyToTensorView(av, "a_in");
+
+  // example-begin fft-1
```
cliffburdick, line 106: You either need to remove this name or give it a unique name since it's colliding with the other example names

include/matx/transforms/fft.h
nvjonwong: Let me correct the comment. Test helped verify.

test/00_transform/FFT.cu
```diff
@@ -92,6 +92,50 @@ TYPED_TEST(FFTTestComplexTypes, FFT1D1024C2C)
...
+  MATX_ENTER_HANDLER();
+  const index_t fft_dim = 1024;
+  this->pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "fft_operators", "fft_1d_fwd", {fft_dim, fft_dim});
+
+  tensor_t<TypeParam, 1> av{{fft_dim}};
+  tensor_t<TypeParam, 1> avo{{fft_dim}};
+  this->pb->NumpyToTensorView(av, "a_in");
+
+  // example-begin fft-1-fwd
```
cliffburdick, line 106: If you add a new sample it should be added to the rst file in the docs:

https://github.com/NVIDIA/MatX/blob/main/docs_input/api/dft/fft/fft.rst

Bonus points for writing a description above it.

test/00_transform/FFT.cu
```diff
@@ -92,6 +92,50 @@ TYPED_TEST(FFTTestComplexTypes, FFT1D1024C2C)
...
+  MATX_ENTER_HANDLER();
+  const index_t fft_dim = 1024;
+  this->pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "fft_operators", "fft_1d_ortho", {fft_dim, fft_dim});
+
+  tensor_t<TypeParam, 1> av{{fft_dim}};
+  tensor_t<TypeParam, 1> avo{{fft_dim}};
+  this->pb->NumpyToTensorView(av, "a_in");
+
+  // example-begin fft-1
```
cliffburdick, line 128: I think these should go in increasing order instead of being called fft-1. We already have up to fft-5:

https://github.com/NVIDIA/MatX/blob/main/docs_input/api/dft/fft/fft.rst

so these would be 6/7

cliffburdick (approved):

cliffburdick (approved):

PR #459, opened 2023-07-31
Add initial polyphase channelizer operator

tbensonatl:
The polyphase channelizer separates an input signal into a set of channels using a provided filter (typically converting a wideband signal at a high sample rate to multiple narrowband signals are lower per-channel sampling rates). This initial version only supports maximally decimated (or critically sampled) cases.


examples/channelize_poly_bench.cu
```diff
@@ -0,0 +1,143 @@
...
+// length, input signal length, and channel range) will be adjusted to a range of interest
+// and the benchmark will be run with and without the proposed kernel changes.
+
+constexpr int NUM_WARMUP_ITERATIONS = 2;
+
+// Number of iterations per timed test. Iteration times are averaged in the report.
+constexpr int NUM_ITERATIONS = 20;
+
+template <typename InType, typename OutType>
+void ChannelizePolyBench(matx::index_t channel_start, matx::index_t channel_stop)
```
cliffburdick, line 55: It's fine here too, but we should probably have an accompanying example in the benchmarks directory for real benchmarking

examples/channelize_poly_bench.cu
```diff
@@ -0,0 +1,143 @@
...
+      const matx::index_t output_len_per_channel = (input_len + num_channels - 1) / num_channels;
+
+      auto input = matx::make_tensor<InType, 2>({num_batches, input_len});
+      auto filter = matx::make_tensor<InType, 1>({filter_len});
+      auto output = matx::make_tensor<OutType, 3>({num_batches, output_len_per_channel, num_channels});
+
+      input.PrefetchDevice(stream);
+      filter.PrefetchDevice(stream);
+      output.PrefetchDevice(stream);
+      cudaStreamSynchronize(stream);
```
cliffburdick, line 91: You should be able to remove these since you have the warmup below

examples/channelize_poly_bench.cu
```diff
@@ -0,0 +1,143 @@
...
+      for (int k = 0; k < NUM_WARMUP_ITERATIONS; k++) {
+        (output = channelize_poly(input, filter, num_channels, decimation_factor)).run(stream);
+      }
+
+      cudaStreamSynchronize(stream);
+
+      double total_elapsed_ms = 0.0;
+      float iter_elapsed_ms = 0.0f;
+      for (int k = 0; k < NUM_ITERATIONS; k++) {
+        cudaEventRecord(start);
```
cliffburdick, line 104: Would it be better to have the events/sync outside the loop?

examples/channelize_poly_bench.cu
tbensonatl: I considered putting it in bench/, but at least for now I mostly used this example "interactively" during development (i.e., I change parameters and data types for most runs).

include/matx/kernels/channelize_poly.cuh
```diff
@@ -0,0 +1,440 @@
...
+
+    auto indims = BlockToIdx(input, blockIdx.z, 1);
+    auto outdims = BlockToIdx(output, blockIdx.z, 2);
+    outdims[ChannelRank] = channel;
+
+    // For C2C transforms, we use ifft() to get the correct sign on the imaginary
+    // components and thus avoid a post-fft conjugation. However, ifft scales by
+    // 1/N, so we pre-scale here by N to compensate. For R2C transforms, we must
+    // use fft() and we handle conjugation in the subsequent unpacking.
+    const auto scale = [num_channels]() -> auto {
```
cliffburdick, line 108: Jonathan changed this recently to add different types of normalization. It seems maybe you want a new type of "none" so you don't have IFFT scale at all?

tbensonatl: Yes, if we can avoid scaling after the ifft, then this scaling can be removed. It looks like we can accomplish that now by setting the FFTNorm mode to FORWARD in the ifft, although NONE may better capture the intent. I can update the PR to do either one (add a NONE scaling mode, or just specify forward for the ifft).

include/matx/operators/channelize_poly.h
```diff
@@ -0,0 +1,143 @@
...
+#include "matx/transforms/channelize_poly.h"
+
+namespace matx {
+
+namespace detail {
+  template<typename OpA, typename FilterType>
+  class ChannelizePolyOp : public BaseOp<ChannelizePolyOp<OpA, FilterType>>
+  {
+    private:
+      using out_t = std::conditional_t<is_complex_v<typename OpA::scalar_type>,
```
cliffburdick, line 46: I think this will break for half, right? How about `complex_from_scalar_t`?

tbensonatl: Good point - updated. I still need to enable matxHalf support for the operator as well. We should at least be able to support half for power-of-two channel counts.

include/matx/kernels/channelize_poly.cuh
tbensonatl: Updated to use FFTNorm::FORWARD with ifft and remove the scaling from the kernel.

examples/channelize_poly_bench.cu
tbensonatl: Yep - I lifted the events out of the loop and removed the prefetch in an update.

cliffburdick (approved):

PR #472, opened 2023-08-28
Add overlap operator

cliffburdick:
Added `overlap` operator for counterpart to tensor `OverlapView`

### DIFF
--- a/docs_input/api/manipulation/rearranging/overlap.rst
+++ b/docs_input/api/manipulation/rearranging/overlap.rst
@@ -1,0 +1,33 @@
+.. _overlap_func:
+
+overlap
+#######
+
+Create an overlapping view an of input operator giving a higher-rank view of the input
+
+For example, the following 1D tensor [1 2 3 4 5] could be cloned into a 2d tensor with a
+window size of 2 and overlap of 1, resulting in:
+ [1 2
+  2 3
+  3 4
+  4 5]
+
+Currently this only works on 1D tensors going to 2D, but may be expanded
+for higher dimensions in the future. Note that if the window size does not
+divide evenly into the existing column dimension, the view may chop off the
+end of the data to make the tensor rectangular.
+
+.. note::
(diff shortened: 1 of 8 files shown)


docs_input/api/manipulation/rearranging/overlap.rst
```diff
@@ -0,0 +1,33 @@
+.. _overlap_func:
+
+overlap
+#######
+
+Create an overlapping view an of input operator giving a higher-rank view of the input
+
+For example, the following 1D tensor [1 2 3 4 5] could be cloned into a 2d tensor with a
+window size of 2 and overlap of 1, resulting in:
+ [1 2
```
tmartin-gh, line 10: Doesn't render the rows correctly.

test/00_operators/OperatorTests.cu
```diff
@@ -2378,6 +2378,84 @@ TYPED_TEST(OperatorTestsNumericAllExecs, Reshape)
...
+  using TestType = std::tuple_element_t<0, TypeParam>;
+  using ExecType = std::tuple_element_t<1, TypeParam>;
+  using inner_type = typename inner_op_type_t<TestType>::type;
+
+  ExecType exec{}; 
+
+
+  tensor_t<TestType, 1> a{{10}};
+  a.SetVals({0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+  auto ao = overlap(a, {4}, {2});
```
tmartin-gh, line 2394: Perhaps also have an ASSERT_EQ on the Shape of ao?  If there is a code error and ao had a larger shape than tested by the double-for loop below, it wouldn't be caught.

docs_input/api/manipulation/rearranging/overlap.rst
cliffburdick: This is an rst file, so it should just be newlines as it's shown there

tbensonatl (approved):

Issue #279, opened 2022-09-27
[FEA] Add support for pwelch()


cliffburdick:
pwelch() is commonly used in signal processing for visualizing spectrum without preprocessing first. This feature request is for preliminary support to match basic use cases of scipy/cusignal's `pwelch()`

PR #479, opened 2023-09-01
Add pwelch operator

tmartin-gh:
This PR addresses https://github.com/NVIDIA/MatX/issues/279

### DIFF
--- a/examples/pwelch.cu
+++ b/examples/pwelch.cu
@@ -1,0 +1,102 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2023, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its
+//    contributors may be used to endorse or promote products derived from
+//    this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+/////////////////////////////////////////////////////////////////////////////////
+
+#include "matx.h"
+#include <cassert>
+#include <cstdio>
+#include <cuda/std/ccomplex>
+
+using namespace matx;
+
+/**
+ * PWelch Example
+ *
+ * This example shows how to estimate the power spectral density of a signal using the pwelch() operator
+ * using Welch's method.
+ *
+ */
+
+int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
+{
+  MATX_ENTER_HANDLER();
+  using complex = cuda::std::complex<float>;
+
+  float exec_time_ms;
+  const int num_iterations = 100;
+  index_t signal_size = 256;
+  index_t nperseg = 32;
+  index_t nfft = nperseg;
+  index_t noverlap = 8;
+  float ftone = 3.0;
+  cudaStream_t stream;
+  cudaStreamCreate(&stream);
+  cudaEvent_t start, stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+
+  // Create input signal as a complex exponential
+  auto sample_index_range = range<0>({signal_size},0.f,1.f);
+  auto phase = 2.f * static_cast<float>(M_PI) * ftone * sample_index_range / static_cast<float>(nfft);
+  auto tmp_x = expj(phase);
+  auto x = make_tensor<complex>({signal_size});
+  (x = tmp_x).run(stream); // pre-compute x, tmp_x is otherwise lazily evaluated
+
+  // Create output tensor
+  auto Pxx  = make_tensor<typename complex::value_type>({nfft});
+
+  // Run one time to pre-cache the FFT plan
+  (Pxx = pwelch(x, nperseg, noverlap, nfft)).run(stream);
+  cudaStreamSynchronize(stream);
+
+  // Start the timing
+  cudaEventRecord(start, stream);
+
+  // Start the timing
+  cudaEventRecord(start, stream);
+
+  for (int iteration = 0; iteration < num_iterations; iteration++) {
+    // Use the PWelch operator
+    (Pxx = pwelch(x, nperseg, noverlap, nfft)).run(stream);
+  }
+
+  cudaEventRecord(stop, stream);
+  cudaStreamSynchronize(stream);
+  cudaEventElapsedTime(&exec_time_ms, start, stop);
+
+  printf("Output Pxx:\n");
+  print(Pxx);
+  printf("PWelchOp avg runtime = %.3f ms\n", exec_time_ms / num_iterations);
+
+  CUDA_CHECK_LAST_ERROR();
+  MATX_EXIT_HANDLER();
+  return 0;
+}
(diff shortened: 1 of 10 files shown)


examples/pwelch.cu
```diff
@@ -0,0 +1,105 @@
...
+  auto Pxx  = make_tensor<typename complex::value_type>({nfft});
+
+  for (index_t k = 0; k < signal_size; k++) {
+    // Create the time domain signal as a complex exponential
+    float phase = static_cast<float>(2.0*M_PI*ftone*static_cast<float>(k)/nfft);
+    x(k) = {static_cast<float>(cos(phase)), static_cast<float>(sin(phase))};
+  }
+
+  // Prefetch the data we just created
+  x.PrefetchDevice(0);
```
cliffburdick, line 79: In the future we can leave these out of examples since they don't really add anything to it

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,139 @@
...
+
+
+#include "matx/core/type_utils.h"
+#include "matx/operators/base_operator.h"
+#include "matx/transforms/pwelch.h"
+
+namespace matx
+{
+  namespace detail {
+    template <typename Op_x>
```
cliffburdick, line 43: Just a minor nit, but all other operators use the convention `OpX`

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,139 @@
...
+    class PWelchOp : public BaseOp<PWelchOp<Op_x>>
+    {
+      private:
+        Op_x x_;
+
+        int nperseg_;
+        int noverlap_;
+        int nfft_;
```
cliffburdick, line 51: Should these be `index_t`?

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,139 @@
...
+        using pwelch_xform_op = bool;
+
+        __MATX_INLINE__ std::string str() const {
+          return "pwelch(" + get_type_str(x_) + ")";
+        }
+
+        __MATX_INLINE__ PWelchOp(const Op_x &x, int nperseg, int noverlap, int nfft) :
+              x_(x), nperseg_(nperseg), noverlap_(noverlap), nfft_(nfft) {
+
+          MATX_ASSERT_STR(x.Rank() == 1, matxInvalidDim, "pwelch:  Only input rank of 1 is supported presently");
```
cliffburdick, line 68: This can use `MATX_STATIC_ASSERT_STR` to do it at compile-time

include/matx/transforms/pwelch.h
```diff
@@ -0,0 +1,83 @@
...
+    __MATX_INLINE__ void pwelch_impl(PxxType Pxx, xType x, int nperseg, int noverlap, int nfft, cudaStream_t stream=0)
+    {
+      MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
+
+      MATX_ASSERT_STR(Pxx.Rank() == x.Rank(), matxInvalidDim, "pwelch:  Pxx rank must be the same as x rank");
+      MATX_ASSERT_STR(nfft >= nperseg, matxInvalidDim, "pwelch:  nfft must be >= nperseg");
+      MATX_ASSERT_STR((noverlap >= 0) && (noverlap < nperseg), matxInvalidDim, "pwelch:  Must have 0 <= noverlap < nperseg");
+
+      // Create overlapping view
+      auto x_with_overlaps = x.OverlapView({nfft}, {noverlap});
```
cliffburdick, line 66: This should probably be `overlap` now so you can take in operators and tensors both

include/matx/transforms/pwelch.h
```diff
@@ -0,0 +1,83 @@
...
+
+      // Create temporary space for fft outputs
+      index_t batches = x_with_overlaps.Shape()[0];
+      auto X_with_overlaps = make_tensor<cuda::std::complex<typename PxxType::scalar_type>>({batches,static_cast<index_t>(nfft)});
```
cliffburdick, line 70: The last two parameters of `make_tensor` here should be `MATX_ASYNC_DEVICE_MEMORY, stream` so it asynchronously allocates the tensor instead of blocking

include/matx/transforms/pwelch.h
```diff
@@ -0,0 +1,83 @@
...
+
+      (X_with_overlaps = fft(x_with_overlaps,nfft)).run(stream);
+
+      // Compute magnitude squared in-place
+      (X_with_overlaps = conj(X_with_overlaps) * X_with_overlaps).run(stream);
+      auto mag_sq_X_with_overlaps = X_with_overlaps.RealView();
+
+      // Perform the reduction across 'batches' rows and normalize
+      float norm_factor = 1.f / static_cast<float>(batches);
```
cliffburdick, line 79: May not be necessary, but `norm_factor` could be cast to `inner_type_t` so it's a double * double or float * float.

test/00_transform/PWelch.cu
```diff
@@ -0,0 +1,101 @@
...
+{
+  MATX_ENTER_HANDLER();
+  test.pb->template InitAndRunTVGenerator<TypeParam>(
+      "00_transforms", "pwelch_operators", "pwelch_complex_exponential", {test.params.signal_size, test.params.nperseg, test.params.noverlap, test.params.nfft, test.params.ftone, test.params.sigma});
+
+  tensor_t<TypeParam, 1> x{{test.params.signal_size}};
+  test.pb->NumpyToTensorView(x, "x_in");
+
+  auto Pxx  = make_tensor<typename TypeParam::value_type>({test.params.nfft});
+  (Pxx = pwelch(x, test.params.nperseg, test.params.noverlap, test.params.nfft)).run();
```
cliffburdick, line 82: Can you add:
```
// example-begin pwelch-test-1
 auto Pxx  = make_tensor<typename TypeParam::value_type>({test.params.nfft});
  (Pxx = pwelch(x, test.params.nperseg, test.params.noverlap, test.params.nfft)).run();
  // example-end pwelch-test-1
```

cliffburdick:
Nice! Can you also add the accompanying rst file for documentation? It should have a `pwelch.rst` probably in docs_input/api/signalimage/general unless you think there's a better place. You can copy an existing rst file, then have your example pointed to: `pwelch-test-1`

cliffburdick (changes requested):
Nice! Can you also add an rst in the appropriate directory in docs_input/api pointing to `pwelch-test-1`?

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,139 @@
...
+
+          Exec(std::make_tuple(tmp_out_), std::forward<Executor>(ex));
+        }
+    };
+  }
+
+  /**
+   *  Estimate the power spectral density of a 1D tensor using Welch's method [1].
+   *
+   * @param x
```
cliffburdick, line 115: Missing `@tparam xType`

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,139 @@
...
+   *   Input time domain tensor x
+   * @param nperseg
+   *   Length of each segment
+   * @param stream
```
cliffburdick, line 119: Stream isn't a parameter here it looks like

include/matx/transforms/pwelch.h
```diff
@@ -0,0 +1,83 @@
...
+/////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+namespace matx
+{
+  /**
+   *  Estimate the power spectral density of a 1D tensor using Welch's method [1].
+   *
+   * @param Pxx
```
cliffburdick, line 40: Missing `tparam` on `PxxType` and `xType`

examples/pwelch.cu
```diff
@@ -0,0 +1,101 @@
...
+  cudaStreamCreate(&stream);
+  cudaEvent_t start, stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+
+  // Create buffers
+  auto x  = make_tensor<complex>({signal_size});
+  auto Pxx  = make_tensor<typename complex::value_type>({nfft});
+
+  for (index_t k = 0; k < signal_size; k++) {
```
luitjens, line 70: your 2.0 should be 2.0f.
 
 we should also consider an operator implementation.  I don't think we quite have what we need here though.  ideally we would have a sincos operator which takes a real and outputs complex.   @cliffburdick for thoughts.

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,137 @@
...
+        index_t noverlap_;
+        index_t nfft_;
+        std::array<index_t, 2> out_dims_;
+        mutable matx::tensor_t<typename OpX::scalar_type, 2> tmp_out_;
+
+      public:
+        using matxop = bool;
+        using scalar_type = typename OpX::scalar_type;
+        using matx_transform_op = bool;
+        using pwelch_xform_op = bool;
```
luitjens, line 59: Why is this necessary?

include/matx/operators/pwelch.h
```diff
@@ -0,0 +1,137 @@
...
+
+        __MATX_INLINE__ std::string str() const {
+          return "pwelch(" + get_type_str(x_) + ")";
+        }
+
+        __MATX_INLINE__ PWelchOp(const OpX &x, index_t nperseg, index_t noverlap, index_t nfft) :
+              x_(x), nperseg_(nperseg), noverlap_(noverlap), nfft_(nfft) {
+
+          MATX_STATIC_ASSERT_STR(x.Rank() == 1, matxInvalidDim, "pwelch:  Only input rank of 1 is supported presently");
+          for (int r = 0; r < x.Rank(); r++) {
```
luitjens, line 69: do we need to store out_dims or can we just query x_ for outdims instead?

examples/pwelch.cu
cliffburdick: If you use 2.0f you should also cast M_PI to cast to a float or it's going to do double math anyways. This is just the example setup, so I'm not sure it matters.

Since this is the initialization code, I also don't have a strong opinion on changing this to an operator. If you wanted to, you could make a `sincosf` and `sincospif`  and pass in a `phase` vector to compute it.

include/matx/operators/pwelch.h
cliffburdick: This was a placeholder to use in places where we wanted to detect specific transform operators. It's not used yet, but could come in handy with type traits later for fusion.

examples/pwelch.cu
tmartin-gh: I can make the 2.0f and M_PI casting change now.  Do you want to include a complex exponential generator operator in this PR, or leave that for a separate PR?

tmartin-gh: Pushed a new commit using the existing expj() operator...

cliffburdick (approved):

PR #481, opened 2023-09-06
Docs cleanup.  Enforce warning-as-error for doxygen and sphinx.

tmartin-gh:


docs_input/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set(DOXYGEN_OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR}/doxygen)
 set(DOXYGEN_INDEX_FILE ${DOXYGEN_OUTPUT_DIR}/html/index.html)
 set(DOXYGEN_EXAMPLE_DIR ${PROJECT_SOURCE_DIR}/examples)
 set(DOXYGEN_EXCLUDE_DIR ${PROJECT_SOURCE_DIR}/third_party/*)
-set(DOXYGEN_EXCLUDE_DIR2 ${PROJECT_BINARY_DIR}/*)
+set(DOXYGEN_EXCLUDE_DIR2 ${PROJECT_SOURCE_DIR}/build*)
```
cliffburdick, line 13: What was the difference here? Wouldn't the old way be more accurate?

cliffburdick (approved):

docs_input/CMakeLists.txt
tmartin-gh: The CMakeLists.txt makes building documentation an orthogonal process from building everything else, so you need to have two build directories.  I used this method to exclude all the possible build directories that might exist in the local repo's filesystem, not just the docs build directory.  Otherwise, if you have an example/test/etc build directory, building the docs will recursively parse through it and take a long time.

PR #491, opened 2023-09-27
Add window operator input to pwelch

tmartin-gh:
This PR adds support for using window functions with the pwelch operator.

### DIFF
--- a/test/00_operators/PWelch.cu
+++ b/test/00_operators/PWelch.cu
@@ -36,21 +36,29 @@
 #include "utilities.h"
 #include "gtest/gtest.h"
 
+#include <pybind11/pybind11.h>
+using namespace pybind11::literals; // to bring in the '_a' literal
+
 using namespace matx;
 
 struct TestParams {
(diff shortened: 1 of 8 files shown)


test/00_operators/PWelch.cu
```diff
@@ -74,16 +81,45 @@ template <typename TypeParam>
...
+  if (test.params.window_name == "none")
+  {
+    (Pxx = pwelch(x, test.params.nperseg, test.params.noverlap, test.params.nfft)).run();
+  }
+  else
+  {
+    auto w = make_tensor<typename TypeParam::value_type>({test.params.nperseg});
+    if (test.params.window_name == "boxcar")
+    {
+      (w = ones<typename TypeParam::value_type>({test.params.nperseg})).run();
```
cliffburdick, line 111: Just FYI `ones` is a dimensionless operator, and you should be able to just write `ones()` and it will adapt to whatever shape it needs to.

cliffburdick (approved):

PR #512, opened 2023-11-02
Optimize resample poly kernels

tbensonatl:
Add additional resample_poly kernels that are more optimal in specific cases (e.g., in the case of a pure downsampler). Currently,
there are three kernels, but only two of them are in use. There are some cases for which the third is fastest, but a pattern for when that will be the case is not yet clear.

### DIFF
--- a/include/matx/kernels/resample_poly.cuh
+++ b/include/matx/kernels/resample_poly.cuh
@@ -173,6 +173,8 @@ __global__ void ResamplePoly1D_PhaseBlock(OutType output, InType input, FilterTy
         }
     }
 
+    __syncthreads();
+
     // left_h_ind is the index in s_filter that contains the filter tap that will be applied to the
     // last input signal value not to the right of the output index in the virtual upsampled array.
     // If the filter has odd length and a given output value aligns with an input value, then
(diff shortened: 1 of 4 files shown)


include/matx/kernels/resample_poly.cuh
```diff
@@ -40,24 +40,38 @@
...
 #include "matx/core/tensor_utils.h"
 
 namespace matx {
 
+// Use for __launch_bounds__ to allow the compiler to tune register usage
+static constexpr int MATX_RESAMPLE_POLY_MAX_NUM_THREADS = 256;
+
+// We use a static 11 KiB buffer to potentially store the filter. If it fits, we will load it
+// into smem. If not, then we will load it from global memory at the time of use. We choose
+// 11 KiB so that we can definitely fit four blocks in 48 KiB, leaving 1 KiB per block
```
cliffburdick, line 60: Have you experimented with dynamic shm and increasing this over L1 slicing?

include/matx/kernels/resample_poly.cuh
```diff
@@ -217,6 +229,254 @@ __global__ void ResamplePoly1D(OutType output, InType input, FilterType filter,
...
+    // All but the last dim are batch indices
+    const int batch_idx = blockIdx.x;
+    auto bdims = BlockToIdx(output, batch_idx, 1);
+
+    // Scale the filter coefficients by up to match scipy's convention
+    const filter_t scale = static_cast<filter_t>(up);
+    const index_t max_input_ind = input_len - 1;
+
+    const index_t filter_len_half = filter_len/2;
+    const index_t filter_central_tap = (filter_len-1)/2;
```
cliffburdick, line 292: Isn't this just filter_len/2 or do you need to be on the left side for even lengths?

cliffburdick (approved):

include/matx/kernels/resample_poly.cuh
tbensonatl: Do you mean adjusting the smem/L1 split via carveout? I have not tried increasing the shared memory allotments, but overall the improvements that I have seen by using shared memory have been relatively modest. If we have cases with relatively long filters (either due to large upsampling/downsampling factors, or due to strict filter performance requirements) then it's worth exploring. The kernel that is implemented but now unused decomposes the problem by phase, so only needs to store the taps for a single phase in smem. That would be ~1400 taps for FP64, so we may just be able to use that existing kernel in such cases without increasing the carveout.

include/matx/transforms/resample_poly.h
```diff
@@ -55,36 +55,115 @@ inline void matxResamplePoly1DInternal(OutType &o, const InType &i,
...
+
+  // We default to the ElemBlock kernel as it tends to work well for general problems.
+  enum class ResampleKernel {
+    PhaseBlock,
+    ElemBlock,
+    WarpCentric,
+  } kernel = ResampleKernel::ElemBlock;
+
+  // The WarpCentric kernel currently uses cg::reduce(), which requires trivially-copyable types.
+  if constexpr (std::is_trivially_copyable_v<output_t>) {
```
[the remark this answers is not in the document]
tbensonatl: I did find it in this section of the docs where it lists the requirements for val:

include/matx/kernels/resample_poly.cuh
tbensonatl: Yes, this is to handle even-lengths. The logic in the subsequent loops effectively assumes an odd-length filter (and thus a central tap). In the case of storing the filter to shared memory, a zero is prepended to the filter (prior to flipping for convolution) to make the length odd. Thus, for an even-length filter not stored to shared memory, (filter_len-1)/2 always references the same coefficient as filter_len/2 for filters stored to shared memory. I had to double-check, so I will update the PR to include a comment here.

PR #529, opened 2023-12-08
New rapids and nvbench

cliffburdick:


tylera-nvidia (approved):

tylera-nvidia:
One thing I noticed is that the docs currently don't have a version # in it for our third-party requirements here: https://nvidia.github.io/MatX/build.html#required-third-party-dependencies

For some customers we may want to list the most-recent versions, as they will need to get those approved to their labs.

also is [libcudacxx](https://github.com/NVIDIA/libcudacxx) still the correct requirement, or should it be updated to CCCL?

cliffburdick:
> One thing I noticed is that the docs currently don't have a version # in it for our third-party requirements here: https://nvidia.github.io/MatX/build.html#required-third-party-dependencies
> 
> For some customers we may want to list the currently used versions, as they will need to get those approved to their labs.
> 
> also is [libcudacxx](https://github.com/NVIDIA/libcudacxx) still the correct requirement, or should it be updated to CCCL?

These are all resolved in the latest commit

PR #541, opened 2023-12-19
Fix batched cov transform

tbensonatl:
Fix a few issues with batched cov() transforms:

- Use transpose_matrix instead of transpose for the batched transposed deviations. Thus, dimensions [0, RANK-2) match the sizes of the batched dimensions from the non-transposed deviations.
- Use stream-oriented async allocators for temporaries and set the memory space to MATX_ASYNC_DEVICE_MEMORY.
- Make some changes in dimension-checking for batched matmul() transforms. Compute the number of batches using the dimensions of C and, if necessary, broadcast A or B into the batched dimensions.
- Add a cov() unit test case with batching and rectangular inputs.

One remaining issue is that we allocate 3-4 times the size of the input for temporary storage. This storage is part of a cached plan that currently is not evicted/freed, so if the dimensions of cov() inputs are changing over time, the cache will grow quickly.

### DIFF
--- a/docs_input/api/manipulation/rearranging/transpose.rst
+++ b/docs_input/api/manipulation/rearranging/transpose.rst
@@ -12,7 +12,7 @@ than calling `permute()` since it's not lazily evaluated and can use an optimize
 Examples
 ~~~~~~~~
 
-.. literalinclude:: ../../../../include/matx/transforms/cov.h
+.. literalinclude:: ../../../../test/00_operators/OperatorTests.cu
    :language: cpp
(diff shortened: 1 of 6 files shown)


cliffburdick (approved):

cliffburdick:
@tbensonatl looks like an example was removed:
```
Warning, treated as error:
/home/jenkins/workspace/unit-tests/docs_input/api/manipulation/rearranging/transpose.rst:15:start-after pattern not found: example-begin transpose-test-1
```

PR #542, opened 2024-01-03
Update caching for transforms to fixing all leaks reported by compute-sanitizer

cliffburdick:
Leaks in pybind are still reported, but are considered innocuous at this point since it's only in unit tests.

### DIFF
--- a/include/matx/core/allocator.h
+++ b/include/matx/core/allocator.h
@@ -83,12 +83,143 @@ struct matxPointerAttr_t {
 inline detail::matxMemoryStats_t matxMemoryStats; ///< Statistics object
 inline std::shared_mutex memory_mtx; ///< Mutex protecting updates from map
 
+struct MemTracker {
+  std::unordered_map<void *, detail::matxPointerAttr_t> allocationMap;
+
+  auto size() {
+    return allocationMap.size();
+  }
+
+  auto deallocate(void *ptr) {
+    MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
+
+    std::unique_lock lck(memory_mtx);
+    auto iter = allocationMap.find(ptr);
+
+    if (iter == allocationMap.end()) {
+  #ifdef MATX_DISABLE_MEM_TRACK_CHECK
+      // This error can occur in situations where the user includes MatX in multiple translation units
+      // and a deallocation occurs in a different one than it was allocated. Allow the user to ignore
+      // these cases if they know the issue.
+      MATX_THROW(matxInvalidParameter, "Couldn't find pointer in allocation cache");
+  #endif    
+      return;
+    }
+
+    size_t bytes = iter->second.size;
+    matxMemoryStats.currentBytesAllocated -= bytes;
+
+    switch (iter->second.kind) {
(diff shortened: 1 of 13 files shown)


include/matx/core/allocator.h
```diff
@@ -83,12 +83,140 @@ struct matxPointerAttr_t {
...
+    case MATX_ASYNC_DEVICE_MEMORY:
+      err = cudaMallocAsync(ptr, bytes, stream);
+      MATX_ASSERT(err == cudaSuccess, matxOutOfMemory);
+      break;
+    case MATX_INVALID_MEMORY:
+      MATX_THROW(matxInvalidType, "Invalid memory kind when allocating!");
+      break;
+    };
+
+    MATX_ASSERT(ptr != nullptr, matxOutOfMemory);
```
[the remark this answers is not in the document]
cliffburdick: I'll change it since that will also prevent it from being removed in release mode

include/matx/transforms/fft.h
```diff
@@ -1065,16 +1061,18 @@ __MATX_INLINE__ void ifft2_impl(OutputTensor o, const InputTensor i,
...
-  if (ret == std::nullopt) {
-    auto tmp = new detail::matxFFTPlan2D_t<decltype(in), decltype(out)>{out, in, stream};
-    detail::cache_2d.Insert(params, static_cast<void *>(tmp));
-    tmp->Inverse(out, in, stream);
-  }
-  else {
-    auto fft_type = static_cast<detail::matxFFTPlan2D_t<decltype(in), decltype(out)> *>(ret.value());
-    fft_type->Inverse(out, in, stream);
-  }
+  using cache_val_type = detail::matxFFTPlan2D_t<decltype(out), decltype(in)>;
```
tylera-nvidia, line 1064: did the out/in order get flipped here? versus the old detail::matxFFTPlan2D, the order has changed for the template parameters

include/matx/transforms/solver.h
```diff
@@ -1529,21 +1525,21 @@ void eig_impl(OutputTensor &&out, WTensor &&w,
...
+      return std::make_shared<cache_val_type>(w_new, tv, jobz, uplo);
+    },
+    [&](std::shared_ptr<cache_val_type> ctype) {
+      ctype->Exec(tv, w_new, tv, jobz, uplo, stream);
+    }
+  );
 
   /* Copy and free async buffer for transpose */
   matx::copy(out, tv.PermuteMatrix(), stream);
+  matxFree(tp);
```
tylera-nvidia, line 1542: tp is used in tv, which is part of the construction of the plan eig_cache_t. do we only use tv for sizing , or is the op itself retained in any way? I just want to make sure we're not freeing memory here that the plan retains in some way.

include/matx/transforms/fft.h
cliffburdick: The previous one was a bug. out, in was correct

include/matx/transforms/solver.h
cliffburdick: It's async-allocated memory, so it will only be freed after the copy to `out`

tbensonatl (approved):

PR #551, opened 2024-01-08
Update docs

tmartin-gh:
Removed the date/time string addition to the version field.  Added a version.rst file that contains some useful version info.


cliffburdick (changes requested):
I think given the size of the commit it would be better to find a time to rewrite the history to remove the directory docs/ instead so we don't have hundreds of files with unnecessary history

PR #559, opened 2024-01-11
outer() operator for outer product

cliffburdick:


include/matx/operators/outer.h
```diff
@@ -0,0 +1,160 @@
...
+        using outer_xform_op = bool;
+
+        __MATX_INLINE__ std::string str() const { 
+          return "outer(" + get_type_str(a_) + "," + get_type_str(b_)  + ")";
+        }
+
+        __MATX_INLINE__ OuterOp(const OpA &A, const OpB &B, float alpha, float beta) : 
+              a_(A), b_(B), alpha_(alpha), beta_(beta) {
+          
+          for (int r = 0; r < Rank(); r++) {
```
luitjens, line 68: is this correct?  I would expect an outer product to increase the rank by 1?   ie:  matrix = outer(vec,vec)

include/matx/transforms/outer.h
```diff
@@ -0,0 +1,103 @@
...
+  MATX_ASSERT_STR(C.Size(C.Rank() - 1) == B.Size(B.Rank() - 1), matxInvalidSize,
+      "outer: last dimension of C must match last dimension of B");
+
+  std::array<index_t, TensorTypeA::Rank() + 1> ac;
+  std::array<index_t, TensorTypeB::Rank() + 1> bc;
+
+  ac.fill(matxKeepDim);
+  bc.fill(matxKeepDim);
+
+  for (int r = 0; r < A.Rank()-1; r++) {
```
luitjens, line 90: is this correct if A is broadcasted and B is not?

cliffburdick: Had this there before broadcasting was supported. Should be fixed.

include/matx/operators/outer.h
cliffburdick: Fixed using max

PR #563, opened 2024-01-12
SVD & QR improvements

luitjens:


examples/svd_power.cu
```diff
@@ -41,19 +41,21 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
 {
   MATX_ENTER_HANDLER();
 
+#if 1
```
cliffburdick, line 44: is this needed?

include/matx/operators/svd.h
```diff
@@ -191,7 +192,7 @@ namespace detail {
       using svd_xform_op = bool;
 
       __MATX_INLINE__ std::string str() const { return "svdpi(" + get_type_str(a_) + ")"; }
-      __MATX_INLINE__ SVDBPIOp(const OpA &a, int iterations) : a_(a), iterations_(iterations)
+      __MATX_INLINE__ SVDBPIOp(const OpA &a, int max_iters=10, float tol = 0.0f) : a_(a), max_iters_(max_iters), tol_(tol)
```
cliffburdick, line 195: I think the default should only be in the user-facing interface since it would get carried down to here

cliffburdick (approved):

examples/svd_power.cu
luitjens: not needed but I use it to toggle between batched and non-batched which has been shown to be useful for debugging/testing.  We can build examples in 20 seconds where as matx_test is 20 minutes.

include/matx/operators/svd.h
luitjens: fixed

PR #568, opened 2024-01-22
Add abs2() operator for squared abs()

tbensonatl:
Add abs2() operator that is equivalent to abs()*abs() without the extra loads and the unnecessary sqrt operators associated with complex abs() calculations.

### DIFF
--- a/docs_input/api/math/misc/abs2.rst
+++ b/docs_input/api/math/misc/abs2.rst
@@ -4,8 +4,8 @@ abs2
 ====
 
 Squared absolute value. For complex numbers, this is the squared
-complex magnitude, or real(t)^2 + imag(t)^2. For real numbers,
-this is equivalent to the squared value, or t*t.
+complex magnitude, or real(t)\ :sup:`2` + imag(t)\ :sup:`2`. For real numbers,
+this is equivalent to the squared value, or t\ :sup:`2`.
 
 .. doxygenfunction:: abs2(Op t) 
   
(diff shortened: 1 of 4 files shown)


docs_input/api/math/misc/abs2.rst
```diff
@@ -0,0 +1,20 @@
+.. _abs2_func:
+
+abs2
+====
+
+Squared absolute value. For complex numbers, this is the squared
+complex magnitude, or real(t)^2 + imag(t)^2. For real numbers,
```
cliffburdick, line 7: If you want to get fancy in rst you can use `\ :sup:`2``

test/00_operators/OperatorTests.cu
```diff
@@ -1582,6 +1582,73 @@ TYPED_TEST(OperatorTestsAllExecs, OperatorFuncs)
...
+TYPED_TEST(OperatorTestsNumericAllExecs, Abs2)
+{
+  MATX_ENTER_HANDLER();
+  using TestType = std::tuple_element_t<0, TypeParam>;
+  using ExecType = std::tuple_element_t<1, TypeParam>;
+  using inner_type = typename inner_op_type_t<TestType>::type;
+
+  ExecType exec{};
+
+  auto sync = [&exec]() constexpr {
```
cliffburdick, line 1594: We should probably have a sync function in the executor at this point

cliffburdick (approved):

test/00_operators/OperatorTests.cu
tbensonatl: Agreed. I can create a PR to add that. I think we only use this lambda sync pattern in the unit tests currently.

PR #576, opened 2024-02-10
Added ARM FFT Support

cliffburdick:
This PR adds initial support for FFTs on ARM using NVPL. NVPL follows the FFTW interface, which prevents a few features cuFFT has. Most notable half-precision support is missing and FFTW requires the data pointer to be part of the plan. This makes caching somewhat pointless since the same size FFT with different pointers cannot reuse a plan. 

Included in the PR is full unit test coverage for all FFT tests that had fp32/fp64 support. Once the NVPL API evolves to possible allow plans without pointers we can rewrite this to use the cuFFT method of having plans based on 1D/2D classes.  

You may also notice a seemingly unrelated change in conv2d. This is an nvcc compiler bug that only affects ARM. This change will be reverted once the compiler bug is addressed

This commit also moves all unit tests to the tuple format so we can switch executors per test quickly.

### DIFF
--- a/include/matx/transforms/fft/fft_fftw.h
+++ b/include/matx/transforms/fft/fft_fftw.h
@@ -1,0 +1,556 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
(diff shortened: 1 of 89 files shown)


include/matx/transforms/fft/fft_fftw.h
```diff
@@ -0,0 +1,545 @@
...
+                                      params.odist,  
+                                      FFTW_ESTIMATE);
+        } 
+        
+        fftw_execute(plan);
+        fftw_destroy_plan(plan);
+        fftw_cleanup();           
+      }
+
+      return plan;
```
tbensonatl, line 360: The plan has been destroyed above; do we want to return it? The return appears unused. Do we have caching for NVPL plans?

cliffburdick: Thanks. This was done prior to knowing nvpl fft's limitations to where having a plan stick around is not beneficial.

luitjens: should we return null instead then?

cliffburdick: yes, it's a void lambda now

tbensonatl (approved):

PR #579, opened 2024-02-15
stream memory cleanup

cliffburdick:
Use the default stream when tearing down caches on static shutdown. 
Allow user to override stream for allocation


include/matx/core/allocator.h
```diff
@@ -334,6 +345,25 @@ inline void matxFree(void *ptr)
...
+{
+  MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
+  
+  GetAllocMap().update_stream(ptr, stream);
+}
+
+inline void matxFree(void *ptr, cudaStream_t stream)
+{
+  MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
+  update_stream(ptr, stream);
```
luitjens, line 363: why not just take a stream into deallocate as an optional second parameter?  No need to have the update_stream API.

cliffburdick: since `cudaStream_t` is just a raw pointer, you cannot use a default parameter. If your default is nullptr that's the default stream, but we want a "no user-provided stream" as the option, which isn't possible since it's indistinguishable from any stream.

luitjens: i wasn't clear in what I was saying.  optional second parameter would not be a default parameter but an overload.

luitjens (approved):

PR #587, opened 2024-03-11
Fix bugs in 2D FFTs and add tests

benbarsdell:
- Fixes several issues where input/output, dimensions, or types were swapped in internal calculations.
- Adds tests for non-square, batched, strided batched, R2C, and C2R 2D transforms.


cliffburdick (approved):

cliffburdick:
Thanks @benbarsdell!

Issue #606, opened 2024-04-11
[BUG]: #define print replaces print functions in other libraries


mfzmullen:
**Describe the Bug**
In source file `/include/matx/core/tensor_utils.h` on [line 1176](https://github.com/NVIDIA/MatX/blob/f402035a32859ad2ac490fd7711de05c2e4392d9/include/matx/core/tensor_utils.h#L1176), the macro does not acknowledge the namespace it is nested in. Any code written or included after this file has been included has its `print` function replaced with MatX's version, even if the other library has a properly namespaced print method written (e.g. not as a macro). This makes the include order of headers affect the compilation of even trivial files, example below.

**To Reproduce**
1. Attempt to compile the code below, using the attached [CMakeLists.txt](https://github.com/NVIDIA/MatX/files/14948064/CMakeLists.txt) (note: I pass in a non-system path with -DCMAKE_PREFIX_PATH for the location of MatX)
2. Compilation fails in Eigen, which has a Eigen::internal::print method, that is replaced by MatX's print. 
3. Compilation and execution succeed if Eigen is included before MatX.

**Expected Behavior**
Compilation and execution success should be independent of header include order.

**Code Snippets**
matxTest.cu
```
#include <matx.h>
#include <Eigen/Core>
#include <iostream>

int main( int argc, char** argv )
{
    std::cout << "Pure CUDA" << std::endl;
}
```

**System Details (please complete the following information):**
 - OS: Ubuntu 22.04
 - CUDA version: 12.2.140
 - g++ version: 11.4.0

**Additional Context**
Not a `find` or `awk` guru, but if you run `find . -type f -exec awk '(/print/ && !/fprint/ && !/printf/ && !/_print_/){print FILENAME ":\t" $0;}' '{}' \;` inside MatX/include/, there seem to be no uses of this print macro in the core library. It is used in examples, tests, and docs though it appears. I think renaming the macro to be matx specific, like matx_print would be sufficient to prevent conflicts with other libraries, and am happy to open a PR when a consensus is reached on how to best fix the issue.

mfzmullen:
I should add, I explicitly specified my CUDA arch in CMakeLists.txt, that will likely need to be changed for anyone trying to reproduce this.

tmartin-gh:
Thank you for the feedback!  This is fixed by PR #607.

mfzmullen:
Nice thanks for the quick fix!

PR #607, opened 2024-04-11
Change print macro to matx namespaced function

tmartin-gh:
Issue #606 correctly points out that use of a "print" macro can conflict with other libraries, such as Eigen.  This PR changes the print macro back to a matx::print function.


cliffburdick:
Fixes #606

PR #613, opened 2024-04-29
Add smem-based polyphase channelizer kernel

tbensonatl:
Add a new polyphase channelizer kernel that utilizes shared memory for both the filter coefficients and input samples. This kernel is automatically selected in cases where the corresponding data will fit in shared memory.

### DIFF
--- a/include/matx/kernels/channelize_poly.cuh
+++ b/include/matx/kernels/channelize_poly.cuh
@@ -174,6 +174,148 @@ __global__ void ChannelizePoly1D(OutType output, InType input, FilterType filter
     }
 }
 
+// This kernel works in cases where the full filter (with potentially some zero padding) and
+// the inputs required to compute elems_per_channel_per_cta outputs all fit into shared memory.
+template <typename OutType, typename InType, typename FilterType>
+__global__ void ChannelizePoly1D_Smem(OutType output, InType input, FilterType filter, index_t elems_per_channel_per_cta)
+{
+    using output_t = typename OutType::scalar_type;
+    using input_t = typename InType::scalar_type;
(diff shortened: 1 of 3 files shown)


include/matx/kernels/channelize_poly.cuh
```diff
@@ -174,6 +174,148 @@ __global__ void ChannelizePoly1D(OutType output, InType input, FilterType filter
...
+    // implicitly padded with zeros. We assume that the kernel selection logic
+    // considers the size of the zero-padded filter since that is what we actually
+    // store in shared memory.
+    const uint32_t num_channels = static_cast<uint32_t>(output.Size(ChannelRank));
+    const uint32_t filter_full_len = static_cast<uint32_t>(filter.Size(0));
+    const uint32_t filter_phase_len = static_cast<uint32_t>((filter_full_len + num_channels - 1) / num_channels);
+
+    filter_t *smem_h = reinterpret_cast<filter_t *>(smem_dyn_align16);
+    size_t smem_input_offset = sizeof(filter_t) * filter_phase_len * num_channels;
+    if (smem_input_offset % sizeof(input_t)) {
```
cliffburdick, line 206: We should probably have a generic function to do this somewhere. Convolution also has the dynamic shmem problem where it needs multiple variables to share the same carveout, and they have different alignments.

cliffburdick (approved):

include/matx/kernels/channelize_poly.cuh
tbensonatl: Agreed. I can check a few other libraries and see if there is a common solution to this already.

PR #619, opened 2024-05-10
Implement csqrt

tylera-nvidia:
Add a complex sqrt function equivalent to the C++ functionality of [csqrt](https://en.cppreference.com/w/c/numeric/complex/csqrt), 

The high-level functionality supported is to always generate a complex return, even if the input is only real. I've come across situations where users sqrt negative data, with the expectation of getting complex from real.

### DIFF
--- a/include/matx/operators/scalar_ops.h
+++ b/include/matx/operators/scalar_ops.h
@@ -173,22 +173,8 @@ template <typename T> struct SqrtF {
 template <typename T>
 static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_csqrt(T v1)
 {
-  if constexpr (is_cuda_complex_v<T>) 
-  {
-    return cuda::std::sqrt(v1);
-  }
-  else {
-    ///\todo TYLER_TODO should probably protect for only float types
-    return sqrt(static_cast<cuda::std::complex<T>>(v1));
-  }
-  if constexpr (!is_cuda_complex_v<T>) {
-    ///\todo TYLER_TODO should probably protect for only float types
-    return sqrt(static_cast<cuda::std::complex<T>>(v1));
-  }
-  else {
-    return cuda::std::sqrt(v1);
-  }  
-
+  static_assert(std::is_floating_point_v<T>, "csqrt() only supports non-complex floating point inputs");
+  return sqrt(static_cast<cuda::std::complex<T>>(v1));
 }
 
 template <typename T> struct CSqrtF {
--- a/include/matx/operators/scalar_ops.h
+++ b/include/matx/operators/scalar_ops.h
@@ -178,9 +178,11 @@ static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_csqrt(T v1)
     return cuda::std::sqrt(v1);
   }
   else {
+    ///\todo TYLER_TODO should probably protect for only float types
     return sqrt(static_cast<cuda::std::complex<T>>(v1));
   }
   if constexpr (!is_cuda_complex_v<T>) {
+    ///\todo TYLER_TODO should probably protect for only float types
     return sqrt(static_cast<cuda::std::complex<T>>(v1));
   }
   else {
--- a/include/matx/operators/scalar_ops.h
+++ b/include/matx/operators/scalar_ops.h
@@ -170,8 +170,36 @@ template <typename T> struct SqrtF {
   }
 };
 
-template <typename T> using SqrtOp = UnOp<T, SqrtF<T>>;
+template <typename T>
+static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_csqrt(T v1)
+{
+  if constexpr (is_cuda_complex_v<T>) 
+  {
+    return cuda::std::sqrt(v1);
+  }
+  else {
+    return sqrt(static_cast<cuda::std::complex<T>>(v1));
+  }
+  if constexpr (!is_cuda_complex_v<T>) {
+    return sqrt(static_cast<cuda::std::complex<T>>(v1));
+  }
+  else {
+    return cuda::std::sqrt(v1);
+  }  
 
+}
+
+template <typename T> struct CSqrtF {
+  static __MATX_INLINE__ std::string str() { return "csqrt"; }
+  static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto op(T v1)
+  {
+    return _internal_csqrt(v1);
+  }
+};
+
+
+template <typename T> using SqrtOp = UnOp<T, SqrtF<T>>;
+template <typename T> using CsqrtOp = UnOp<T, CSqrtF<T>>;
 
 template <typename T>
 static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_conj(T v1)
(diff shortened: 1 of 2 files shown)


tylera-nvidia:
@cliffburdick @luitjens. This is the implementation of `csqrt` we discussed in 2022. There is a C++ implementation of csqrt, however I was not able to get that function to compile in a .cu file, as it doesn't look like nvcc supports it. I was able to get equivalent functionality using a static cast prior to making the sqrt call, however I imagine this does leave us open to a divergent behavior to the C++ `csqrt` if that behavior ever changes. Any thoughts on if that makes this function still "worth it"?

include/matx/operators/scalar_ops.h
```diff
@@ -170,8 +170,38 @@ template <typename T> struct SqrtF {
   }
 };
 
-template <typename T> using SqrtOp = UnOp<T, SqrtF<T>>;
+template <typename T>
+static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_csqrt(T v1)
+{
+  if constexpr (is_cuda_complex_v<T>)
```
cliffburdick, line 176: Do we want to do this or return an error if they pass a complex to this since it doesn't follow C++ libs?

include/matx/operators/scalar_ops.h
```diff
@@ -170,8 +170,38 @@ template <typename T> struct SqrtF {
...
+{
+  if constexpr (is_cuda_complex_v<T>) 
+  {
+    return cuda::std::sqrt(v1);
+  }
+  else {
+    ///\todo TYLER_TODO should probably protect for only float types
+    return sqrt(static_cast<cuda::std::complex<T>>(v1));
+  }
+  if constexpr (!is_cuda_complex_v<T>) {
```
cliffburdick, line 184: This shouldn't be needed anymore. I think we're going to clean these up and dump the old 11.4 requirement

cliffburdick (changes requested):
2 comments left

include/matx/operators/scalar_ops.h
tylera-nvidia: It currently fails to compile, because it can't complete the type; for example if it's an int,  `complex<int>` can't be implemented so it errors. one sec, I'll post the exact error message. I thought it was relatively self-explanatory, but I can add better checks if we want

tylera-nvidia: error generates for trying to feed in an int:

````
/home/tallen/scratch/playground/matxExample/build/_deps/cccl-src/libcudacxx/include/cuda/std/detail/libcxx/include/complex(1382): error: no instance of function template "cuda::std::__4::polar" matches the argument list
            argument types are: (double, int)
      return cuda::std::__4::polar(cuda::std::__4::sqrt(cuda::std::__4::abs(__x)), cuda::std::__4::arg(__x) / _Tp(2));
             ^
          detected during instantiation of "cuda::std::__4::complex<_Tp> cuda::std::__4::sqrt(const cuda::std::__4::complex<_Tp> &) [with _Tp=int]" at line 182 of /home/tallen/scratch/playground/matxExample/MatX/include/matx/operators/scalar_ops.h

1 error detected in the compilation of "/home/tallen/scratch/playground/matxExample/complexSetVals.cu".
make[2]: *** [CMakeFiles/complexSetVals.dir/build.make:76: CMakeFiles/complexSetVals.dir/complexSetVals.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:201: CMakeFiles/complexSetVals.dir/all] Error 2
make: *** [Makefile:91: all] Error 2

Comment (tylera-nvidia) [reply include/matx/operators/scalar_ops.h:176]: so it does not currently point back to the line where the int is being used, but it does complain about int in a complex... so probably could be more helpful

Comment (cliffburdick) [reply include/matx/operators/scalar_ops.h:176]: I thought this function is only for scalars of floating point type. Can this just assert it's a floating point type, and if so, just call the `cuda::std` version and that's it?

Comment (tylera-nvidia) [reply include/matx/operators/scalar_ops.h:176]: this should take in a float type (non complex) and return a complex type of the same float type as the input. Ie:
`float` -> `complex<float>`
`double` -> `complex<double>`

I _thought_ the C++ standard also supported `complex<float>` -> `complex<float>` for it's csqrt, but it looks like the problem is that the `csqrt `function is a c function (C99), and does not support the `std::complex`, so it does not support complex inputs.

so I think the valid types is actually just this:

`
template <typename T>
static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_csqrt(T v1)
std::enable_if_t<std::is_floating_point_v<T>, bool> = true>
{
  return sqrt(static_cast<cuda::std::complex<T>>(v1));
}
`

if you think that makes sense I'll update the code to the above.

Comment (tylera-nvidia) [reply include/matx/operators/scalar_ops.h:176]: we could easily make ours support `complex<float>` as well (it would just call the normal sqrt), but that would go beyond the standard C++ support pattern.

Comment (cliffburdick) [reply include/matx/operators/scalar_ops.h:176]: I would remove the `enable_if` and instead make it a `static_assert` with a message telling the user it's not supported. otherwise they get a bad template error they have to figure out.

Comment (tylera-nvidia) [reply include/matx/operators/scalar_ops.h]: alrighty, updated per our discussion. Tested it all works as expected.
Works with float/double, errors on any other type (example error below for using int)

```
Consolidate compiler generated dependencies of target complexSetVals
[ 20%] Built target cuBlasExample
[ 50%] Building CUDA object CMakeFiles/nvtx_autoGen.dir/nvtx_autoGen.cu.o
[ 50%] Building CUDA object CMakeFiles/vectorMult.dir/vectorMult.cu.o
[ 50%] Building CUDA object CMakeFiles/matxExample.dir/test.cu.o
[ 60%] Building CUDA object CMakeFiles/complexSetVals.dir/complexSetVals.cu.o
/home/tallen/scratch/playground/matxExample/MatX/include/matx/operators/scalar_ops.h(176): error: static assertion failed with "csqrt() only supports non-complex floating point inputs"
    static_assert(std::is_floating_point_v<T>, "csqrt() only supports non-complex floating point inputs");
    ^
          detected during:
            instantiation of "auto matx::detail::_internal_csqrt(T) [with T=int]" at line 184
            instantiation of "auto matx::detail::CSqrtF<T>::op(T) [with T=int]" at line 102
            instantiation of "auto matx::detail::UnOp<T1, F>::op(const T1 &) [with T1=int, F=matx::detail::CSqrtF<int>]" at line 106
            instantiation of class "matx::detail::UnOp<T1, F> [with T1=int, F=matx::detail::CSqrtF<int>]" at line 365 of /home/tallen/scratch/playground/matxExample/MatX/include/matx/operators/unary_operators.h
            instantiation of "auto matx::csqrt(I1) [with I1=matx::tensor_t<int, 1, matx::basic_storage<matx::raw_pointer_buffer<int, matx::matx_allocator<int>>>, matx::tensor_desc_t<std::array<matx::index_t, 1UL>, std::array<matx::index_t, 1UL>, 1>>, <unnamed>=void]" at line 53 of /home/tallen/scratch/playground/matxExample/complexSetVals.cu

1 error detected in the compilation of "/home/tallen/scratch/playground/matxExample/complexSetVals.cu".
make[2]: *** [CMakeFiles/complexSetVals.dir/build.make:76: CMakeFiles/complexSetVals.dir/complexSetVals.cu.o] Error 2
```

include/matx/operators/scalar_ops.h
tylera-nvidia: I believe this is handled with the redesign above. let me know if I'm missing any other condition you want to cover.

cliffburdick (approved):

PR #621, opened 2024-05-14
Fixing Clang errors to compile with clang-17

### LINKED ISSUE #622: [BUG] Clang cannot compile as a host compiler
Clang-17 shows a lot of warnings when compiling matx

cliffburdick:
Closes #622

### DIFF
--- a/include/matx/core/half.h
+++ b/include/matx/core/half.h
@@ -54,14 +54,14 @@ template <typename T> struct alignas(sizeof(T)) matxHalf {
    * @brief Default constructor
    * 
    */
-  __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ matxHalf() = default;
+  __MATX_INLINE__ matxHalf() = default;
 
   /**
    * @brief Default copy constructor
    * 
    * @param x_ Parameter to copy
    */
-  __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ matxHalf(const matxHalf<T> &x_) noexcept = default;
+  __MATX_INLINE__ matxHalf(const matxHalf<T> &x_) noexcept = default;
 
   /**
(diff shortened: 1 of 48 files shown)


include/matx/core/half.h
tylera-nvidia, line 1: do we intend to lose these constructors on the device? I noticed there are more files that you changed from `__host__ __device__` to no qualifier, so want to make sure there isn't some assumption built in here, and that we're actually removing use of them on the device.

include/matx/generators/random.h
```diff
@@ -546,6 +552,9 @@ template <typename T, int RANK> class randomTensorView_t {
             curandGenerateNormalDouble(gen_, &tmp[1], 1, beta_, alpha_);
           }
         }
+        else {
+          val = 0;
```
tylera-nvidia, line 556: Do we want to return zero versus throwing an error? I would assume something is broken if I call a random generator and get all zeros.

include/matx/generators/random.h
```diff
@@ -401,6 +401,11 @@ template <typename T, int RANK> class randomTensorView_t {
         using scalar_type = T;
         using matxop = bool;
 
+        static_assert(std::is_same_v<T, float> ||
```
tylera-nvidia, line 404: We should update the documentation for random to reflect it only works for float types. currently we don't have any verbiage to outline that

include/matx/generators/random.h
cliffburdick: This is device code, so we can't really throw an error here

include/matx/core/half.h
cliffburdick: it was getting an error in clang, and since we don't use these on the device at the moment I'm going to leave them this way. If we have a need to later we can add it back.

tylera-nvidia (approved):

PR #640, opened 2024-05-30
Add mixed precision support for channelize_poly

tbensonatl:
Add support for mixing precisions of the input and filter for channelize_poly.


include/matx/transforms/channelize_poly.h
```diff
@@ -289,19 +292,20 @@ inline void channelize_poly_impl(OutType out, const InType &in, const FilterType
       // slice to create a tensor view of only [0, num_channels-1]. This guarantees that we
       // always stride by an even number of elements from one batch to the next while exposing
       // a tensor view of appropriate dimensions.
+      using post_filter_t = typename inner_op_type_t<output_t>::type;
```
cliffburdick, line 295: I think this is my mistake, but the `_t` form of the type should return ::type already.

cliffburdick (approved):

PR #642, opened 2024-05-31
Basic synchronization support with sync()

aayushg55:
- Added synchronization support through sync() for both host and cuda executors
- Updated all tests to use calls to sync() instead of cudaStreamSynchronize() or cudaDeviceSynchronize()
- Added sync() to the documentation
- Removed calls to Prefetch

Other:
- Fix to Upsample test in OperatorTests.cu to sync before comparison and catch errors
- Fix to CmakeLists.txt file to copy test.npy for an 00_io test

### DIFF
--- a/bench/00_operators/operators.cu
+++ b/bench/00_operators/operators.cu
@@ -10,6 +10,7 @@ template <typename ValueType>
 void vector_add(nvbench::state &state, nvbench::type_list<ValueType>)
 {
   // Get current parameters:
+  cudaExecutor exec{0};
   const int x_len = static_cast<int>(state.get_int64("Vector size"));
 
   state.add_element_count(x_len, "NumElements");
@@ -19,9 +20,9 @@ void vector_add(nvbench::state &state, nvbench::type_list<ValueType>)
 
   tensor_t<ValueType, 1> xv{{x_len}};
   tensor_t<ValueType, 1> xv2{{x_len}};
-  xv.PrefetchDevice(0);
-  (xv = xv + xv2).run();
-  cudaDeviceSynchronize();
+
+  (xv = xv + xv2).run(exec);
+  exec.sync();
 
   state.exec( 
     [&xv, &xv2](nvbench::launch &launch) {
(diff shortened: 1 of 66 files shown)


bench/00_operators/operators.cu
```diff
@@ -19,9 +20,10 @@ void vector_add(nvbench::state &state, nvbench::type_list<ValueType>)
 
   tensor_t<ValueType, 1> xv{{x_len}};
   tensor_t<ValueType, 1> xv2{{x_len}};
+
   xv.PrefetchDevice(0);
```
cliffburdick, line 24: I should have told you this, but can you please remove all the Prefetch calls like this one too?

docs_input/notebooks/exercises/solutions/example2_assignment1.cu
```diff
@@ -78,10 +80,10 @@ int main() {
   /****************************************************************************************************
    * Add 5.0 to all elements of A and store the results back in A
    ****************************************************************************************************/
-  (A = A + 5.0).run();
+  (A = A + 5.0).run(exec);
```
cliffburdick, line 83: Thanks for adding this, but for these notebook examples I would leave out the executor and sync. These are intended for brand new users to learn the syntax, so I don't know if we should introduce syncs when they're learning the syntax. This could be added to a new notebook section, though.

cliffburdick: On second thought, just leave this in

cliffburdick (changes requested):
Nice work! I noticed you fixed several bugs too. Just a few small comments. Please rebase those into a single commit once you update and we can push this through CI

bench/00_operators/operators.cu
aayushg55: What should they be replaced by, if anything?

cliffburdick: Nothing. They can be removed

cliffburdick (approved):

PR #647, opened 2024-06-07
Change check on build type to avoid short circuiting

### LINKED ISSUE #646: [BUG] Configuration error using Ninja Multi-Config.
**Describe the Bug**

Trying to configure the project using `Ninja Multi-Config` generator result in an error.

**To Reproduce**

Doing the following:

```bash
mkdir build && cd build
cmake -G "Ninja Multi-Config" ..
```

Will result in an error:

```bash
CMake Error at CMakeLists.txt:126 (if):
  if given arguments:

    "NOT" "CMAKE_BUILD_TYPE" "OR" "STREQUAL" "Debug"

  Unknown arguments specified
```

**System Details**
 - OS: AlmaLinux release 8.4
 - CUDA version: 12.2
 - g++ version: 11.2

**Additional Context**

The issue comes from the `CMakeLists.txt` file line 126:
```
if (NOT CMAKE_BUILD_TYPE OR ${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
CMake conditions does not use short-circuit as stated [there](https://cmake.org/cmake/help/latest/command/if.html#condition-syntax)

> Binary logical operators AND and OR, from left to right, without any short-circuit.

CMake being CMake, the issue only appears using `Ninja Multi-Config`.

cliffburdick:
Fixes #646

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,8 +123,10 @@ target_link_libraries(matx INTERFACE CCCL::CCCL)
 
 # Set flags for compiling tests faster
 set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --threads 0 -ftemplate-backtrace-limit=0)
-if (NOT CMAKE_BUILD_TYPE OR ${CMAKE_BUILD_TYPE} STREQUAL "Debug")
-    set(MATX_CUDA_FLAGS ${MATX_CUDA_FLAGS} -g -lineinfo)
+
+# Hack because CMake doesn't have short circult evaluation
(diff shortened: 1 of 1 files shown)


raplonu:
@cliffburdick thanks for looking at this issue.

Another solution could be do the following:

```
if (NOT CMAKE_BUILD_TYPE OR "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
```

`"${CMAKE_BUILD_TYPE}"` will still be considered as a `STRING` even if empty.

PR #650, opened 2024-06-13
Add complex cast operators for split inputs

tbensonatl:
Add complex cast operators that cast from split real and imaginary operators to cuda::std::complex

### DIFF
--- a/docs_input/api/casting/as_complex_float.rst
+++ b/docs_input/api/casting/as_complex_float.rst
@@ -1,0 +1,24 @@
+.. _as_complex_float_func:
+
+as_complex_float
+=================
+
+Cast an operator to cuda::std::complex<float>
+
+.. doxygenfunction:: matx::as_complex_float(T t)
+.. doxygenfunction:: matx::as_complex_float(T1 t1, T2 t2)
+
+Examples
+~~~~~~~~
+
+.. literalinclude:: ../../../test/00_operators/OperatorTests.cu
+   :language: cpp
+   :start-after: example-begin as_complex_float-test-1
+   :end-before: example-end as_complex_float-test-1
+   :dedent:
+
+.. literalinclude:: ../../../test/00_operators/OperatorTests.cu
+   :language: cpp
+   :start-after: example-begin as_complex_float-test-2
+   :end-before: example-end as_complex_float-test-2
+   :dedent:
--- a/include/matx/operators/cast.h
+++ b/include/matx/operators/cast.h
@@ -126,10 +126,8 @@ namespace matx
 	      __MATX_INLINE__ std::string str() const { return as_type_str<NewType>() + "(" + real_op_.str() + "," + imag_op_.str() + ")"; }
         __MATX_INLINE__ ComplexCastOp(T1 real_op, T2 imag_op) : real_op_(real_op), imag_op_(imag_op) {
           static_assert(detail::get_rank<T1>() == detail::get_rank<T2>(), "rank of real and imaginary operators must match");
-          for (auto i = 0; i < detail::get_rank<T1>(); i++) {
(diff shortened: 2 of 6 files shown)


docs_input/api/casting/as_complex_float.rst
```diff
@@ -0,0 +1,24 @@
+.. _as_complex_float_func:
+
+as_complex_float
+=================
+
+Cast an operator to cuda::std::complex<float>
+
+.. doxygenfunction:: matx::as_complex_float(T t)
```
cliffburdick, line 8: Do we need these for the half types?

include/matx/operators/cast.h
```diff
@@ -108,8 +108,80 @@ namespace matx
...
+        using scalar_type = NewType;
+        static_assert(!is_complex_v<T1> && !is_complex_half_v<T1>, "T1 input operator cannot be complex");
+        static_assert(!is_complex_v<T2> && !is_complex_half_v<T2>, "T2 input operator cannot be complex");
+        static_assert(is_complex_v<NewType> || is_complex_half_v<NewType>, "ComplexCastOp output type should be complex");
+
+	      __MATX_INLINE__ std::string str() const { return as_type_str<NewType>() + "(" + real_op_.str() + "," + imag_op_.str() + ")"; }
+        __MATX_INLINE__ ComplexCastOp(T1 real_op, T2 imag_op) : real_op_(real_op), imag_op_(imag_op) {
+          static_assert(detail::get_rank<T1>() == detail::get_rank<T2>(), "rank of real and imaginary operators must match");
+          for (auto i = 0; i < detail::get_rank<T1>(); i++) {
+            if (real_op_.Size(i) != imag_op_.Size(i)) {
```
cliffburdick, line 130: I think this can be reduced to 
```
op1.Shape() == op2.Shape()
```
but I haven't tried

include/matx/operators/cast.h
```diff
@@ -108,8 +108,80 @@ namespace matx
...
+              MATX_THROW(matxInvalidSize, "ComplexCastOp: sizes of input operators must match in all dimensions");
+            }
+          }
+        };
+
+        template <typename... Is>
+        __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) operator()(Is... indices) const
```
cliffburdick, line 137: Do we not want `auto` here?

cliffburdick (approved):

docs_input/api/casting/as_complex_float.rst
tbensonatl: Probably, but I'm not sure what the best types would be. matxHalfComplex<matxFp16> and matxHalfComplex<matxBf16>? The cast operator name would need to differentiate between matxFp16 and matxBf16, or pick only one as the return value for as_complex_half.

include/matx/operators/cast.h
tbensonatl: Yes, I believe so. I copy-pasted that from CastOp, so can change both. I'll check what our other `operator()(...) const` return types are as well.

docs_input/api/casting/as_complex_float.rst
cliffburdick: Sure, let's leave it for now and add if someone requests it.

include/matx/operators/cast.h
tbensonatl: I changed this case to auto, but we may want a follow-up PR to revert other operators to auto.

include/matx/operators/cast.h
tbensonatl: Changed to a Shape() comparison and added a test case to verify that the exception is thrown for mismatched sizes.

PR #659, opened 2024-07-01
Added plan caching for FFTW host plans

aayushg55:


include/matx/core/type_utils.h
```diff
@@ -473,12 +473,27 @@ template <> struct is_fp16_type<matxFp16> : std::true_type {};
...
  */
 template <class T> inline constexpr bool is_fp16_type_v = detail::is_fp16_type<T>::value;
 
+/**
+ * @brief Determine if a type is an FP32 type
+ * 
+ * @tparam T Type to test
+ */
+template<typename T>
+inline constexpr bool is_fp32_type_v = std::is_same_v<typename inner_op_type_t<T>::type, float>;
```
cliffburdick, line 488: you can use `std::is_same_v<float, T>` instead of making a new type trait

include/matx/transforms/fft/fft_fftw.h
```diff
@@ -208,6 +220,280 @@ struct FftFFTWparams_t {
...
+      if (init_fp64_) {
+          fftw_cleanup_threads();
+          fftw_cleanup();
+          init_fp64_ = false;
+      }
+    }
+  }
+
+private:
+  static inline int active_plans_ = 0;
```
cliffburdick, line 302: This won't be thread-safe and will lead to a race condition with multiple threads. You can make it a `cuda::std::atomic<int>` to solve that. Same thing for the other variables.

cliffburdick (changes requested):
just a couple small changes

include/matx/core/type_utils.h
aayushg55: I had added this for checking if the type was a complex float, since the following was used a couple of times in fft_fftw.h: `std::is_same_v<typename inner_op_type_t<typename OutputTensor::scalar_type>::type, float>`

include/matx/transforms/fft/fft_fftw.h
aayushg55: Sure, will update

include/matx/core/type_utils.h
cliffburdick: I would change the name to `is_fp32_inner_type` then.

aayushg55: Updated

cliffburdick (approved):

PR #665, opened 2024-07-12
NVPL BLAS Support

aayushg55:
- Integrated NVPL BLAS for host GEMMs
- Enabled MatMul, Outer, and MatVec operators for host
- Updated MatMul.cu unit tests to test the host executor


CMakeLists.txt
```diff
@@ -172,6 +172,20 @@ set(WARN_FLAGS ${WARN_FLAGS} $<$<COMPILE_LANGUAGE:CXX>:-Werror>)
 set (CUTLASS_INC "")
 target_compile_definitions(matx INTERFACE MATX_ENABLE_CUTLASS=0)
 
+if (MATX_NVTX_FLAGS)
+    add_definitions(-DMATX_NVTX_FLAGS)
+    target_compile_definitions(matx INTERFACE MATX_NVTX_FLAGS)
+endif()
+if (MATX_BUILD_32_BIT)
+    set(INT_TYPE "lp64")
```
cliffburdick, line 180: Should this be called `BLAS_INT_TYPE`?

test/00_transform/MatMul.cu
```diff
@@ -101,136 +105,156 @@ TYPED_TEST(MatMulTestFloatTypes, SmallRect)
...
-  this->pb->NumpyToTensorView(b, "b");
-
-  // example-begin matmul-test-1
-  // Perform the GEMM C = A*B
-  (c = matmul(a, b)).run();
-  // example-end matmul-test-1
-  MATX_TEST_ASSERT_COMPARE(this->pb, c, "c", this->thresh);
-
+  using ExecType = cuda::std::tuple_element_t<1, TypeParam>;
+  if constexpr (!detail::CheckMatMulSupport<ExecType, TestType>()) {
```
cliffburdick, line 109: Is this needed every time if it's done in `SetUp`?

cliffburdick (approved):
Nice job! I think the docs are

test/00_transform/MatMul.cu
aayushg55: Without the check in the test itself, it will fail a static assert in matmul_impl when it tries to compile the test with a half precision datatype.

CMakeLists.txt
aayushg55: Other NVPL libraries like LAPACK also follow this convention of 2 interfaces. ArmPL also uses this convention. Should we use some other name?

cliffburdick: Nope, this is fine.

PR #668, opened 2024-07-17
enh: add macro to define cuda functions accessible at global scope

mfzmullen:
Adding a macro to define CUDA intrinsic math functions accessible at global scope, not in `cuda::std` namespace, calling the macro `MATX_GLOBAL_UNARY_OP_GEN`. 

Added the `rsqrt` function for computing reciprocal sqrts using the new macro. When not using the '-use_fast_math' flag, benchmarks using the hyperfine tool show `1.0/matx::sqrt` runs about 1.22x slower than 'matx::rsqrt'. Note when -use_fast_math is enabled, 1.0/sqrtf is replaced with rsqrtf, and the benchmark reflects that in similar run times with 100 runs (and 3 warmup runs).

### DIFF
--- a/docs_input/api/math/misc/rsqrt.rst
+++ b/docs_input/api/math/misc/rsqrt.rst
@@ -1,0 +1,18 @@
+.. _rsqrt_func:
+
+rsqrt
+====
+
+Reciprocal square root
+
+.. doxygenfunction:: rsqrt(Op t) 
+  
+Examples
+~~~~~~~~
+
+.. literalinclude:: ../../../../test/00_operators/OperatorTests.cu
+   :language: cpp
+   :start-after: example-begin rsqrt-test-1
+   :end-before: example-end rsqrt-test-1
+   :dedent:
+
--- a/include/matx/core/half.h
+++ b/include/matx/core/half.h
@@ -689,6 +689,41 @@ sqrt(const matxHalf<__nv_bfloat16> &x)
 }
 
 /**
+ * @brief Reciprocal Square root value
+ * 
+ * @tparam T Type of half
+ * @param x Value of half
+ * @return Reciprocal square root of input
+ */
+template <class T>
+__MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ matxHalf<T> rsqrt(const matxHalf<T> &x)
+{
+#ifdef __CUDA_ARCH__
+  return hrsqrt(x.x);
+#else
+  return static_cast<T>(::rsqrt(static_cast<float>(x.x)));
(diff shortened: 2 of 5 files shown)


cliffburdick (changes requested):
Thanks for the PR. I think rather than make a macro, instead this should be an explicit operator where you specify the implementation, like:

https://github.com/NVIDIA/MatX/blob/main/include/matx/operators/scalar_ops.h#L548

The reason for this is there's an intrinsic that only works on float/double like you found, but there's also an hrsqrt that works on fp16/bf16 that you could add compile-time checks on to call those versions if needed.

cliffburdick:
> Point well taken. Thanks! Could I mimic the `sqrt` function in half.h and then use the following in scalar_ops.h?
> 
> ```
> template <typename T>
> static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto _internal_rsqrt(T v1)
> {
>   if constexpr (is_matx_type_v<T>){
>     return rsqrt(v1);
>   }
>   else {
>     return ::rsqrt(v1);
>   }
> }
> template <typename T> struct RSqrtF {
>   static __MATX_INLINE__ std::string str() { return "rsqrt"; }
>   static __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ auto op(T v1)
>   {
>     return _internal_rsqrt(v1);
>   }
> };
> template <typename T> using RSqrtOp = UnOp<T, RSqrtF<T>>;
> ```
> 
> where the `rsqrt` (not `::rsqrt`) should resolve to the respective implementation in half.h when called with one of the MatX half types?

Yes, that works. Adding it in the half header is fine, but will likely need to change in the future when half types become standard in C++.

cliffburdick:
Also, since this is a new operator it needs a unit test in OperatorTests.cu and documentation added to docs_input. It should be easy to follow existing functions, but let me know if you need help.

mfzmullen:
I can follow along with existing functions, thanks. 

More as an aside, won't everything in half.h need to change when halfs become standard C++?

cliffburdick:
> I can follow along with existing functions, thanks.
> 
> More as an aside, won't everything in half.h need to change when halfs become standard C++?

Yes. Ideally we would get rid of that header entirely since we shouldn't have our own bespoke half types in a header.

mfzmullen:
@cliffburdick is there a setting where after the first failed build, subsequent builds fail if the changes haven't been approved?

cliffburdick (approved):

cliffburdick:
Thanks @mfzmullen !

PR #674, opened 2024-07-19
Fix to ConvCorr tests to skip host tests when host not enabled

aayushg55:


test/00_transform/ConvCorr.cu
```diff
@@ -142,6 +162,19 @@ protected:
   }
 
   void TearDown() override { pb.reset(); }
+
+  void CheckExecSupport() {
```
cliffburdick, line 166: This function can be marked constexpr too

test/00_transform/ConvCorr.cu
```diff
@@ -78,6 +79,19 @@ protected:
...
```
cliffburdick, line 83: `constexpr` function too

test/00_transform/ConvCorr.cu
aayushg55: Fixed

cliffburdick (approved):

PR #675, opened 2024-07-19
Expanded Host BLAS support

aayushg55:
- Added x86 BLAS support with OpenBLAS and BLIS. Supports the OpenMP or pthread versions for multithreading. Used the standard cblas_?gemm api and looped over higher dimensions
- Updated NVPL BLAS calls to get better multithreaded performance. cblas_?gemm_batch is always running single-core so replaced it with cblas_?gemm_batch_strided. (Note: Revert if bug is fixed and original achieves better performance.)


CMakeLists.txt
```diff
@@ -211,6 +215,19 @@ if (MATX_EN_NVPL OR MATX_EN_X86_FFTW)
             target_link_libraries(matx INTERFACE ${FFTW_LIB} ${FFTWF_LIB} ${FFTW_OMP_LIB} ${FFTWF_OMP_LIB})
             target_compile_definitions(matx INTERFACE MATX_EN_X86_FFTW=1)
         endif()
+
+        # BLAS
```
cliffburdick, line 219: Should there be a check to make sure that they don't have multiple BLAS libraries enabled (including NVPL)?

CMakeLists.txt
```diff
@@ -68,6 +70,7 @@ set(MATX_EN_PYBIND11 OFF CACHE BOOL "Enable pybind11 support")
 set(cutensor_DIR "" CACHE PATH "Directory where cuTENSOR is installed.")
 set(cutensornet_DIR "" CACHE PATH "Directory where cuTensorNet is installed.")
 set(eigen_DIR "" CACHE PATH "Directory where Eigen is installed")
+set(blas_DIR "" CACHE PATH "Directory where OpenBLAS or BLIS is installed (install prefix)")
```
cliffburdick, line 73: or NVPL?

cliffburdick (approved):
Nice work!

CMakeLists.txt
aayushg55: Yes, will add

CMakeLists.txt
aayushg55: Updated

CMakeLists.txt
aayushg55: Added

PR #676, opened 2024-07-21
Update README.md

HugoPhibbs:
adding bold for easy to miss adding of matx target. For C++ novices, this can be useful. 

Thx


cliffburdick (approved):
Thanks!

PR #683, opened 2024-07-24
Added toeplitz operator

### LINKED ISSUE #663: [FEA] Add toeplitz operator
Add a `toeplitz` that matches the functionality of SciPy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.toeplitz.html

cliffburdick:
Added `toeplitz` operator
Renamed `scalar_type` to `value_type` in operators to better match STL

Closes #663

### DIFF
--- a/docs_input/api/linalg/other/toeplitz.rst
+++ b/docs_input/api/linalg/other/toeplitz.rst
@@ -6,7 +6,7 @@ toeplitz
 Generate a toeplitz matrix
 
 `c` represents the first column of the matrix while `r` represents the first row. `c` and `r` must
-have the same first value; if they don't match, the first value from `r` will be used.
+have the same first value; if they don't match, the first value from `c` will be used.
 
 Passing a single array/operator as input is equivalent to passing the conjugate of the same
 input as the second parameter. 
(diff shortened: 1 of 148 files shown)


docs_input/api/linalg/other/toeplitz.rst
```diff
@@ -0,0 +1,33 @@
+.. _toeplitz_func:
+
+toeplitz
+========
+
+Generate a toeplitz matrix
+
+`c` represents the first column of the matrix while `r` represents the first row. `c` and `r` must
+have the same first value; if they don't match, the first value from `r` will be used.
```
tylera-nvidia, line 9: We document here that if c and r don't have a first value that match we use r[0], however looking at the code, it looks like we always use op1[0] for the first element, which maps to c param from the user API.

tylera-nvidia: (this is because we use the conditional j>i). We could change the conditional or change the wording here to reflect is always used for (0,0) and its diagonal.

tylera-nvidia (approved):

docs_input/api/linalg/other/toeplitz.rst
cliffburdick: Yes, I updated the code after writing that. Thanks

PR #695, opened 2024-08-05
Add support for UINT Generation 

tylera-nvidia:
Currently only calling the `curandGenerate` function to generate a fully random bit for 32bits, scaling range to full uint32_t size.

Considering adding another randomi function to provide inputs for scaling range (min/max).


include/matx/generators/random.h
```diff
@@ -54,6 +54,28 @@ __global__ void curand_setup_kernel(Gen *states, uint64_t seed, index_t size)
...
+ *
+ * @tparam T inner type of data
+ * @tparam Gen Generator type
+ * @param val Value to store in
+ * @param state Generator state
+ * @param min min of the range to generate
+ * @param max max of the range to generate
+ */
+template <typename T, typename Gen>
+__inline__ __MATX_DEVICE__ void get_randomi(T &val, Gen *state, double min, double max)
```
cliffburdick, line 69: I thought you said curand_uniform returns a number between 0 and FLOAT_MAX, but it looks like it's 0..1?

include/matx/generators/random.h
```diff
@@ -186,233 +206,62 @@ template <typename T> class [[deprecated("Use random() operator instead of rando
...
-        uint64_t seed_;
-        inner_t alpha_;
-        inner_t beta_;
+        uint64_t seed_;     
         bool init_ = false;
         bool device_;
-
+        
+        union{
+        randFloatParams<inner_t> fParams_;
```
cliffburdick, line 244: Looks like the spacing is off here

include/matx/generators/random.h
```diff
@@ -186,233 +206,62 @@ template <typename T> class [[deprecated("Use random() operator instead of rando
...
-  __MATX_INLINE__ void PostRun([[maybe_unused]] ShapeType &&shape, [[maybe_unused]] Executor &&ex) noexcept  
-  {
-  }  
-};
-
-  namespace detail {
+  
+  template< typename T >
+  struct randFloatParams{
+    Distribution_t dist_;
```
cliffburdick, line 225: Isn't this structure only used for normal distributions?

include/matx/generators/random.h
```diff
@@ -186,233 +206,62 @@ template <typename T> class [[deprecated("Use random() operator instead of rando
...
         __MATX_INLINE__ std::string str() const { return "random"; }
 
         // Shapeless constructor to be allocated at run invocation
         RandomOp() = delete;
-
-        inline RandomOp(ShapeType &&s, Distribution_t dist, uint64_t seed, inner_t alpha, inner_t beta) :
-          dist_(dist), seed_(seed), alpha_(alpha), beta_(beta)
+  
+        // base constructor, should never be called directly
+        inline RandomOp(ShapeType &&s, uint64_t seed) :
```
cliffburdick, line 263: Can you change these to `__MATX_INLINE__` instead to be consistent with other files? I know it was like that already.

include/matx/generators/random.h
```diff
@@ -186,233 +206,62 @@ template <typename T> class [[deprecated("Use random() operator instead of rando
...
+  
+  template< typename T >
+  struct randFloatParams{
+    Distribution_t dist_;
+    T alpha_;
+    T beta_;
+  };  
+  
     template <typename T, typename ShapeType>
     class RandomOp : public BaseOp<RandomOp<T, ShapeType>> {
```
cliffburdick, line 231: I was thinking RandomOp would take the distribution as a template parameter, then you don't need to save it inside the structs. Everything would be compile-time.

include/matx/generators/random.h
```diff
@@ -587,13 +497,24 @@ template <typename T, int RANK> class randomTensorView_t {
            std::enable_if_t<!std::is_array_v<remove_cvref_t<ShapeType>>, bool> = true>
   inline auto random(ShapeType &&s, Distribution_t dist, uint64_t seed = 0, LowerType alpha = 1, LowerType beta = 0)
   {
+    static_assert(
```
cliffburdick, line 500: Maybe use the second parameter to add a string saying why this is enforced.

include/matx/generators/random.h
```diff
@@ -611,4 +532,57 @@ template <typename T, int RANK> class randomTensorView_t {
...
+   * @param min min of generation range
+   * @param max max of generation range
+   * @return Random number operator
+   */
+  template <typename T, typename ShapeType, typename LowerType = typename inner_op_type_t<T>::type,
+           std::enable_if_t<!std::is_array_v<remove_cvref_t<ShapeType>>, bool> = true>
+  inline auto randomi(ShapeType &&s, uint64_t seed = 0, LowerType min = 0, LowerType max = 100)
+  {
+    static_assert(
+                  std::is_same_v<T, uint32_t> ||
```
cliffburdick, line 555: Same comment with a string

test/00_tensor/ViewTests.cu
```diff
@@ -469,6 +469,56 @@ TYPED_TEST(ViewTestsFloatNonComplexNonHalf, Random)
...
+    (t3f = (TestType)0).run(this->exec);
+    (t3f = randomi<TestType>({count, count, count}, 0, minBound, maxBound )).run(this->exec);
+    // example-end randomi-test-1   
+    this->exec.sync();
+
+    for (index_t i = 0; i < count; i++) {
+      for (index_t j = 0; j < count; j++) {
+        for (index_t k = 0; k < count; k++) {
+          TestType val = t3f(i, j, k); 
+          ASSERT_NE(val, 0); //bounds are (min,max], so even if min == 0, it shouldn't appear in the data
```
cliffburdick, line 495: I would expect it to be [0,N). What's the reason 0 can't be included?

cliffburdick (changes requested):
Just a few comments

test/00_tensor/ViewTests.cu
tylera-nvidia: This is the distribution of the [curand_uniform](https://docs.nvidia.com/cuda/curand/device-api-overview.html#distributions) API, and since we're using that under the hood, not much we can do about it. I also think it would  be weird to bias the distribution differently than curand natively does, since we say we're using it under the hood.

tylera-nvidia: I need to remove this check anyways though, as it was failing due to the distributions covering 0, so this line and the comment is being removed.

include/matx/generators/random.h
tylera-nvidia: yep, updated all in this file.

include/matx/generators/random.h
tylera-nvidia: fixed, in the way I think you mean. I indented the values inside the union (like it was a scope), and left the union at the level of the parent class as it was. 

the only other union I could find in our code was at a namespace scope, so I mirrored that spacing here. let me know if you want it a different way.

include/matx/generators/random.h
tylera-nvidia: The reason I didn't do this is that only float supports multiple distribution types; integral is always uniform, so you would need to provide a meaningless input for integral types if we make it a template parameter. you could set it to whatever for integral and we just ignore it.

include/matx/generators/random.h
tylera-nvidia: updated

include/matx/generators/random.h
tylera-nvidia: updated

include/matx/generators/random.h
tylera-nvidia: it's used to [scale the output of the uniform path ](https://github.com/NVIDIA/MatX/blob/c0cea227db63949de4f873b4cf2d4da18c536f37/include/matx/generators/random.h#L427)for floats as well as being used for [parameters to the random generation for the normal path](https://github.com/NVIDIA/MatX/blob/c0cea227db63949de4f873b4cf2d4da18c536f37/include/matx/generators/random.h#L430).

cliffburdick (approved):

PR #697, opened 2024-08-06
Implement argminmax

tylera-nvidia:
Implement a new argminmax function based on cub_reduce that calculates the following:

- an index of the data's minimum
- an index of the data's maximum
- the data's max
- the data's min

example call would be:

` (matx::mtie(minVal, minIdx, maxVal, maxIdx) = matx::argminmax(inFlattened)).run();  `


include/matx/generators/random.h
```diff
@@ -518,6 +539,9 @@ template <typename T, int RANK> class randomTensorView_t {
           if constexpr (std::is_same_v<T, float>) {
             curandGenerateUniform(gen_, &val, 1);
           }
+          else if constexpr (std::is_same_v<T, uint32_t>) {
```
cliffburdick, line 542: Doesn't this need a `randomi` function?

tylera-nvidia: this has an older version of the random changes. I just needed a working version for a test function. I'll update this one once I have the other PR completed.

cliffburdick:
@tylera-nvidia this will need a docs page too

include/matx/transforms/cub.h
```diff
@@ -878,6 +938,42 @@ inline void ExecSort(OutputTensor &a_out,
...
+   */
+  inline void ExecArgMinMax(const InputOperator &in,
+                            const cudaStream_t stream)
+  {
+#ifdef __CUDACC__
+    MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
+  
+    // Set up CUB stuff
+    auto d_in = thrust::counting_iterator<size_t>(0);
+    auto min_op = minmaxer{in.Data()};
```
tmartin-gh, line 966: Passing the .Data() pointer will cause this to fail if the underlying tensor has been permuted.  Ideally we'd rework RandomOperatorIterator and RandomOperatorOutputIterator to work correctly with Thrust.

At a minimum, suggest updating the docs (limitations.rst) to indicate this doesn't support permutations.

include/matx/transforms/cub.h
```diff
@@ -1187,15 +1284,99 @@ void cub_reduce(OutputTensor &a_out, const InputOperator &a, typename InputOpera
...
+ *   Sorted tensor
+ * @param a
+ *   Input tensor
+ * @param init
+ *   Value to initialize the reduction with
+ * @param stream
+ *   CUDA stream
+ */
+template <typename OutType, typename TensorIndexType, typename InType>
+void cub_reduce_custom(OutType minDest, TensorIndexType &minIdxs, OutType maxDest, TensorIndexType &maxIdxs, const InType &in, const cudaStream_t stream = 0)
```
tmartin-gh, line 1317: Should this be 'cub_reduce_argminmax' instead of 'cub_reduce_custom' since it doesn't accept an arbitrary custom compare?

tmartin-gh: Or perhaps allow it to accept an arbitrary custom compare, but make this a 'cub_reduce_custom_4output' since it's outputting 4 values?

include/matx/transforms/cub.h
```diff
@@ -1187,15 +1284,99 @@ void cub_reduce(OutputTensor &a_out, const InputOperator &a, typename InputOpera
...
+
+
+/**
+ * Reduce a tensor using CUB but accessing data through compare operator
+ *
+ * Reduces a tensor using the CUB library for either a 0D or 1D output tensor. There
+ * is an existing reduce() implementation as part of matx_reduce.h, but this function
+ * exists in cases where CUB is more optimal/faster.
+ *
+ * @tparam OutputTensor
```
tmartin-gh, line 1303: Doxygen comments don't match code (OutputTensor vs OutType, etc).

include/matx/transforms/cub.h
```diff
@@ -878,6 +938,42 @@ inline void ExecSort(OutputTensor &a_out,
 #endif
   }
 
+  /**
+   * Execute a max on a tensor
+   *
+   * @note Views being passed must be in row-major order
+   *
+   * @tparam OutputTensor
```
tmartin-gh, line 946: Doxygen comments don't match code

cliffburdick:
@tylera-nvidia can we close this one given #778 ?

cliffburdick:
@tylera-nvidia I saw you just committed another patch, but is this OBE given #778?

tylera-nvidia:
@cliffburdick  I was doing some performance comparison with @tmartin-gh  to make sure things were roughly equivalent. I can into some performance issues when the return tensor is size {1} versus size {0}. runtime goes from ~40us to >1ms, but we still get the right answer. My implementation did not exhibit that behavior, but it looks like it was due to my "dumb" pointer arithmetic in writing out the output. 

After resolving that, performance seems roughly comparable, so I'm going to close this PR. It looks like caching is currently broken in main, and we have some unprotected conditions for weird outputs, but those should all be fixed with new branches off main, not based on my old implementation.

PR #716, opened 2024-08-13
Rapids cmake 24 06 package

cliffburdick:
The `make install` method was not working for CCCL since it didn't assign an export set. This PR moves our installation for CCCL to rapids-cmake to handle that for us


CMakeLists.txt
```diff
@@ -118,10 +122,11 @@ if (CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 11.5)
     message(FATAL_ERROR "MatX requires CUDA 11.5 or higher. Please update before using.")
 endif()
 
-# libcudacxx is now part of CCCL, so grab that repo if needed
-message(STATUS "Need CCCL. Finding...")
-set(CCCL_VERSION "v2.4.0" CACHE STRING "Version of CCCL to use")
-include(cmake/FindCCCL.cmake)
+message(STATUS "Finding CCCL...")
+rapids_cpm_cccl(
```
tylera-nvidia, line 126: I saw we updated the rapids_cmake version. we currently don't mention it or its version in the build dependencies [here](https://nvidia.github.io/MatX/build.html#required-third-party-dependencies)... do you think we should? 
Does using the `rapids_cpm_cccl` incur any other dependencies that the cmake `FindCCCL` didn't?

tylera-nvidia: I think we justified _not_ listing the rapids_cmake as a dependency since it's direcly included in the source here.

cliffburdick: It's part of a subtree/our code, so we don't need to mention it since the user won't have to do anything about it

cliffburdick: > I saw we updated the rapids_cmake version. we currently don't mention it or its version in the build dependencies [here](https://nvidia.github.io/MatX/build.html#required-third-party-dependencies)... do you think we should? Does using the `rapids_cpm_cccl` incur any other dependencies that the cmake `FindCCCL` didn't?

All the dependencies are in rapids-cmake itself. We can actually start to move nvbench and google test over to this too and it should handle it much better.

tylera-nvidia (approved):

PR #720, opened 2024-08-16
MATX_EN_CUTENSOR / MATX_ENABLE_CUTENSOR Unified Variable

tylera-nvidia:
Currently we use MATX_EN_CUTENSOR as the cmake flag to control using cuTensor, but use MATX_ENABLE_CUTENSOR for the internal macro define check in code. This causes confusing reporting/error messages, and requires different variables depending how the user is building MatX.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,7 +260,7 @@ if (MATX_EN_CUTENSOR)
 
     include(cmake/FindcuTENSOR.cmake)
     include(cmake/FindcuTensorNet.cmake)
-    target_compile_definitions(matx INTERFACE MATX_ENABLE_CUTENSOR)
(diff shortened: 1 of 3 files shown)


bench/00_transform/einsum.cu
```diff
@@ -1,7 +1,7 @@
 #include <nvbench/nvbench.cuh>
 #include "matx.h"
 
-#if ENABLE_CUTENSOR
+#if MATX_EN_CUTENSOR
```
cliffburdick, line 4: This should be `#ifdef`. Bug was there prior

cliffburdick (approved):

bench/00_transform/einsum.cu
tylera-nvidia: good catch I didn't notice. updated

PR #721, opened 2024-08-16
Solver SVD Optimizations and Improved cuSolver batching

aayushg55:
- Bypassed temp allocations and having to transpose outputs in most cases for SVD
- Added m < n CUDA SVD support (cuSolver natively only supports m >= n)
- Added option to use `gesdd` or `gesvd` LAPACK driver for Host SVD
- For cuSolver batching, replaced per-iteration sync & error check with a single check at the end to reduce blocking operations

### DIFF
--- a/include/matx/transforms/chol/chol_cuda.h
+++ b/include/matx/transforms/chol/chol_cuda.h
@@ -170,7 +170,13 @@ public:
     cudaStreamSynchronize(stream);
 
     for (const auto& info : h_info) {
-      MATX_ASSERT(info == 0, matxSolverError);
+      if (info < 0) {
+        MATX_ASSERT_STR_EXP(info, 0, matxSolverError,
+          ("Parameter " + std::to_string(-info) + " had an illegal value in cuSolver Xpotrf").c_str());
+      } else {
+        MATX_ASSERT_STR_EXP(info, 0, matxSolverError, 
+          (std::to_string(info) + "-th leading minor is not positive definite in cuSolver Xpotrf").c_str());
+      }
     }
   }
 
--- a/include/matx/transforms/chol/chol_cuda.h
+++ b/include/matx/transforms/chol/chol_cuda.h
@@ -147,8 +147,8 @@ public:
       (out = a).run(exec);
     }
 
-    cusolverDnSetStream(this->handle, exec.getStream());
-    int info;
+    const auto stream = exec.getStream();
+    cusolverDnSetStream(this->handle, stream);
 
     // At this time cuSolver does not have a batched 64-bit cholesky interface.
     // Change this to use the batched version once available.
@@ -161,9 +161,15 @@ public:
           this->d_info + i);
 
       MATX_ASSERT(ret == CUSOLVER_STATUS_SUCCESS, matxSolverError);
-      
-      // This will block. Figure this out later
-      cudaMemcpy(&info, this->d_info + i, sizeof(info), cudaMemcpyDeviceToHost);
+    }
+
+    std::vector<int> h_info(this->batch_a_ptrs.size());
+    cudaMemcpyAsync(h_info.data(), this->d_info, sizeof(int) * this->batch_a_ptrs.size(), cudaMemcpyDeviceToHost, stream);
+    
+    // This will block. Figure this out later
+    cudaStreamSynchronize(stream);
+
+    for (const auto& info : h_info) {
       MATX_ASSERT(info == 0, matxSolverError);
     }
   }
--- a/include/matx/transforms/eig/eig_cuda.h
+++ b/include/matx/transforms/eig/eig_cuda.h
@@ -197,7 +197,13 @@ public:
     cudaStreamSynchronize(stream);
 
     for (const auto& info : h_info) {
-      MATX_ASSERT(info == 0, matxSolverError);
+      if (info < 0) {
+        MATX_ASSERT_STR_EXP(info, 0, matxSolverError,
+          ("Parameter " + std::to_string(-info) + " had an illegal value in cuSolver Xsyevd").c_str());
+      } else {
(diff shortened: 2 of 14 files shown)


include/matx/transforms/chol/chol_cuda.h
```diff
@@ -161,9 +161,15 @@ class matxDnCholCUDAPlan_t : matxDnCUDASolver_t {
...
+    }
+
+    std::vector<int> h_info(this->batch_a_ptrs.size());
+    cudaMemcpyAsync(h_info.data(), this->d_info, sizeof(int) * this->batch_a_ptrs.size(), cudaMemcpyDeviceToHost, stream);
+    
+    // This will block. Figure this out later
+    cudaStreamSynchronize(stream);
+
+    for (const auto& info : h_info) {
       MATX_ASSERT(info == 0, matxSolverError);
```
cliffburdick, line 173: While you're here it might be good to add a more descriptive string onto this. The docs say: 
```
l. If info = -i, the i-th parameter is wrong (not counting handle). If info = i, the leading minor of order i is not positive definite.
```

include/matx/transforms/eig/eig_cuda.h
```diff
@@ -188,9 +188,15 @@ class matxDnEigCUDAPlan_t : matxDnCUDASolver_t {
...
+    std::vector<int> h_info(this->batch_a_ptrs.size());
+    cudaMemcpyAsync(h_info.data(), this->d_info, sizeof(int) * this->batch_a_ptrs.size(), cudaMemcpyDeviceToHost, stream);
+
+    // This will block. Figure this out later
+    cudaStreamSynchronize(stream);
 
-      // This will block. Figure this out later
-      cudaMemcpy(&info, this->d_info + i, sizeof(info), cudaMemcpyDeviceToHost);
+    for (const auto& info : h_info) {
       MATX_ASSERT(info == 0, matxSolverError);
```
cliffburdick, line 200: Same as the comment above

include/matx/transforms/lu/lu_cuda.h
```diff
@@ -170,9 +170,15 @@ class matxDnLUCUDAPlan_t : matxDnCUDASolver_t {
           this->d_info + i);
 
       MATX_ASSERT(ret == CUSOLVER_STATUS_SUCCESS, matxSolverError);
```
cliffburdick, line 172: Can you change these to use `MATX_ASSERT_STR_EXP`? This will tell the user the cuSolver error code so it might help debug

include/matx/transforms/chol/chol_cuda.h
aayushg55: Sure, I'll update these and the rest (also the LAPACK error checks)

include/matx/transforms/svd/svd_cuda.h
```diff
@@ -755,72 +756,115 @@ using svd_cuda_cache_t = std::unordered_map<DnSVDCUDAParams_t, std::any, DnSVDCU
...
-  );
+    if(!vt_new.isSameView(vt)) {
+      (vt = vt_new).run(exec);
+    }
+  } else {
+    // get col-major A
+    auto tv = TransposeCopy(tp, a, exec);
+    auto tvt = tv.PermuteMatrix();
+
+    auto u_col_maj = make_tensor<T1>(u.Shape(), MATX_ASYNC_DEVICE_MEMORY, stream);
```
cliffburdick, line 840: Above it uses `exec` to make a copy, but it seems like here it's assuming it's a stream. I think it's ok since we're in the CUDA file, but just making sure that's expected.

include/matx/transforms/svd/svd_lapack.h
```diff
@@ -130,50 +130,80 @@ class matxDnSVDHostPlan_t : matxDnHostSolver_t<typename ATensor::value_type> {
...
+
+      MATX_ASSERT(info == 0, matxSolverError);
+
+      this->liwork = 8 * mn; // iwork has size 8*min(M,N) and is used for all types
+
+      // the real part of the first elem of work holds the optimal lwork.
+      if constexpr (is_complex_v<T1>) {
+        this->lwork = static_cast<lapack_int_t>(work_query.real());
+        
+        lapack_int_t mnthr = (mn * 5) / 3;
```
cliffburdick, line 177: What is the 5/3?

include/matx/transforms/svd/svd_lapack.h
```diff
@@ -130,50 +130,80 @@ class matxDnSVDHostPlan_t : matxDnHostSolver_t<typename ATensor::value_type> {
...
+        this->lrwork = 0;
+      }
+    } else if (params.algo == SVDHostAlgo::DC) {
+      gesdd_dispatch("A", &params.m, &params.n, nullptr, &params.m, nullptr,
+                    nullptr, &params.m, nullptr, &params.n, &work_query,
+                    &this->lwork, nullptr, nullptr, &info);
+
+      MATX_ASSERT(info == 0, matxSolverError);
+
+      this->liwork = 8 * mn; // iwork has size 8*min(M,N) and is used for all types
```
cliffburdick, line 171: Are all these constants documented in the NVPL/LAPACk docs?

aayushg55: These are from the LAPACK [docs](https://netlib.org/lapack/explore-html-3.6.1/d4/d84/group__complex_g_esing_gabe88d0f182417696a6e15a417911dae9.html). All of the libraries follow the same workspace requirements:

include/matx/transforms/svd/svd_cuda.h
aayushg55: Yea, I try to use `exec` when possible but `make_tensor` only supports a stream argument currently.

cliffburdick (approved):

PR #724, opened 2024-08-19
Fix rank check on reductions

### LINKED ISSUE #723: [BUG] sum function produces incorrect results
**Describe the Bug**
I'm using the sum function over a 2D tensor, and instead of doing a reduction along the cols. It doesn't seem to be adding values together properly.

**To Reproduce**

Basically I run the code:

```cpp
template<typename T>
inline int *constructQueryVectorDegreeArrayMatx(matx::tensor_t<T, 2> &distances, T eps) {
    std::cout << eps << std::endl;
    print(distances);

    auto lt = distances < eps;
    auto lt_int = matx::as_type<int>(lt);

    print(lt_int);
    auto res = matx::make_tensor<int>({1, distances.Shape()[1]}, matx::MATX_MANAGED_MEMORY);
    (res = matx::sum(lt_int, {0})).run();

    print(res);
    return res.Data();
}
```

With input/output:
```
2.1
Tensor{float} Rank: 2, Sizes:[4, 4], Strides:[4,1]
000000:  0.0000e+00  1.0000e+00  2.0000e+00  3.0000e+00 
000001:  0.0000e+00  2.0000e+00  1.0000e+00  0.0000e+00 
000002:  1.0000e+00  8.0000e+00  9.0000e+00  1.1000e+01 
000003:  1.5000e+01  2.0000e+00  6.0000e+00  7.0000e+00 
Operator{int32_t} Rank: 2, Sizes:[4, 4]
000000:  1  1  1  0 
000001:  1  1  1  1 
000002:  1  0  0  0 
000003:  0  1  0  0 
Tensor{int32_t} Rank: 2, Sizes:[1, 4], Strides:[4,1]
000000:  9  9  9  9 // This should be different
```

**Expected Behavior**
You would expect `res` to be:

```
{3, 4, 1, 1}
```

But it instead full of 9s  - which is coincidentally the _total_ sum of the expected result.

**System Details (please complete the following information):**
 - OS: Ubuntu 20.04
 - CUDA version: 12.6
 - g++ version: 9.4.0

cliffburdick:
Closes #723

### DIFF
--- a/include/matx/operators/all.h
+++ b/include/matx/operators/all.h
@@ -135,7 +135,7 @@ namespace detail {
 template <typename InType, int D>
 __MATX_INLINE__ auto all(const InType &in, const int (&dims)[D])
 {
-  static_assert(D < InType::Rank(), "reduction dimensions must be <= Rank of input");
+  static_assert(D < InType::Rank(), "reduction dimensions must be < Rank of input");
   auto perm = detail::getPermuteDims<InType::Rank()>(dims);
   auto permop = permute(in, perm);
 
--- a/include/matx/operators/any.h
+++ b/include/matx/operators/any.h
@@ -135,7 +135,7 @@ namespace detail {
 template <typename InType, int D>
 __MATX_INLINE__ auto any(const InType &in, const int (&dims)[D])
 {
-  static_assert(D < InType::Rank(), "reduction dimensions must be <= Rank of input");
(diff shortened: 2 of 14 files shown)


luitjens (approved):

luitjens (changes requested):
This doesn't look correct.  Reduction with D == input rank is fine as we would reduce to a rank0.

PR #726, opened 2024-08-20
mtie should output the correct rank and size for the output operator.

### LINKED ISSUE #725: [BUG] Transform Ops have incorrect rank and size
When a user creates a transform op then queries the rank and size of that op it returns -1 for rank and 0 for size.   

The cause of this is that we return an mtie for transform ops here:

https://github.com/NVIDIA/MatX/blob/main/include/matx/core/tensor.h#L285-L296

Mtie currently returns matxNoRank for rank and 0 for size here:

https://github.com/NVIDIA/MatX/blob/main/include/matx/core/tie.h#L84-L92

This causes issues if you try to then pass that transform into other ops.

luitjens:
Fixes #725

### DIFF
--- a/test/00_transform/Norm.cu
+++ b/test/00_transform/Norm.cu
@@ -85,10 +85,22 @@ TYPED_TEST(NormTestFloatTypes, VectorL1)
   this->pb->RunTVGenerator("vector_l1");
   this->pb->NumpyToTensorView(this->in_v, "in_v");
   this->pb->NumpyToTensorView(this->out_v, "out_v");
+
+  auto redOp = vector_norm(this->in_v, NormOrder::L1);
+
+  EXPECT_TRUE(redOp.Rank() == this->out_v.Rank());
+  for(int i = 0; i < redOp.Rank(); i++) {
+    EXPECT_TRUE(redOp.Size(i) == this->out_v.Size(i));
+  }
+  (this->out_v = redOp).run(this->exec);
+  MATX_TEST_ASSERT_COMPARE(this->pb, this->out_v, "out_v", this->thresh);
+
+  (this->out_v = TestType(0)).run(this->exec);
+
   // example-begin vector-norm-test-1
   (this->out_v = vector_norm(this->in_v, NormOrder::L1)).run(this->exec);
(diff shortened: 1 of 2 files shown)


test/00_transform/Norm.cu
```diff
@@ -85,8 +85,16 @@ TYPED_TEST(NormTestFloatTypes, VectorL1)
...
+  auto redOp = vector_norm(this->in_v, NormOrder::L1);
+
+  EXPECT_TRUE(redOp.Rank() == this->out_v.Rank());
+  for(int i = 0; i < redOp.Rank(); i++) {
+    EXPECT_TRUE(redOp.Size(i) == this->out_v.Size(i));
+  }
+
   // example-begin vector-norm-test-1
-  (this->out_v = vector_norm(this->in_v, NormOrder::L1)).run(this->exec);
+  (this->out_v = redOp).run(this->exec);
```
cliffburdick, line 97: This changes the example in the docs so it no longer shows the norm. This should probably be a separate test without a docs example

test/00_transform/Norm.cu
```diff
@@ -102,8 +110,15 @@ TYPED_TEST(NormTestFloatTypes, VectorL2)
...
+  
+  auto redOp = vector_norm(this->in_v, NormOrder::L2);
+  
+  EXPECT_TRUE(redOp.Rank() == this->out_v.Rank());
+  for(int i = 0; i < redOp.Rank(); i++) {
+    EXPECT_TRUE(redOp.Size(i) == this->out_v.Size(i));
+  }
   // example-begin vector-norm-test-2
-  (this->out_v = vector_norm(this->in_v, NormOrder::L2)).run(this->exec);
+  (this->out_v = redOp).run(this->exec);
```
cliffburdick, line 121: Same

test/00_transform/Norm.cu
luitjens: fixed

test/00_transform/Norm.cu
luitjens: fixed

cliffburdick (approved):

PR #735, opened 2024-08-23
Add some single threaded algorithms

tmartin-gh:
Single threaded algorithms are useful building blocks for running many small problems, one per thread, within a fused kernel.

This PR adds the following single threaded algorithms plus an example program showing their use.
   - SVD (power iteration method)
   - Gauss Newton nonlinear least squares solver
   - Trust Region Reflective nonlinear least squares solver
   - Basic linear algebra utilities to help implement the above (matmul, norm, etc)

TODO:
  - Add unit tests

### DIFF
--- a/docs_input/api/index.rst
+++ b/docs_input/api/index.rst
@@ -10,7 +10,7 @@ API Reference
    manipulation/index.rst
    io/index.rst
    math/index.rst
-   linalg/index.rst   
+   linalg/index.rst
    stats/index.rst
    searchsort/index.rst
    logic/index.rst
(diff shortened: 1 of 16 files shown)


cliffburdick (changes requested):
All the code looks good.

I'm wondering how this fits into the project in general. These are all implementations of algorithms we either do or don't support at the moment. It looks like users would not use these algorithms directly? Maybe instead of exposing the user to these functions as the examples show, they should instead be behind a switch inside the operator itself? For example, this version may be faster than cuSolver when the number of batches is very large, so we would choose to execute this algorithm instead of cuSolver. Similarly on the ones we don't have support for right now, we can call this one by default and if/when we add a parallel version later we can choose to call that one.

PR #740, opened 2024-08-28
Added new pinv() operator and updated Reduced SVD

### LINKED ISSUE #487: [FEA] Add pseudo-inverse operator
Add a (Moore-Penrose) matrix pseudo-inverse operator similar to matlab or scipy pinv(). Pseudo-inverses are used in various problems where a true matrix inverse isn’t available, one example being a least squares fit.

aayushg55:
- Created new `pinv()` operator for matrix pseudo-inverse using SVD in Reduced mode. Optional `rcond` parameter which is used to define the cutoff for small singular values. Singular values smaller than `rcond * largest_singular_value` are set to 0.
- Changed `SVDMode::REDUCED` interface so that U and VT tensors passed in must be of reduced size instead of the full size. That is, instead of  U being of shape `... x m x m`, it should have shape `... x m x k` where `k=min(m,n)`. Similarly VT should be of shape `... x k x n`.
- Added `det()` documentation page

Resolves #487

### DIFF
--- a/include/matx/core/tensor.h
+++ b/include/matx/core/tensor.h
@@ -658,9 +658,10 @@ public:
   {
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
+    [[maybe_unused]] stride_type prod = std::accumulate(std::begin(shape), std::end(shape), 1, std::multiplies<stride_type>());
     // Ensure new shape's total size is not larger than the original
     MATX_ASSERT_STR(
-        sizeof(M) * shape.TotalSize() <= storage_.Bytes(), matxInvalidSize,
+        sizeof(M) * prod <= storage_.Bytes(), matxInvalidSize,
         "Total size of new tensor must not be larger than the original");
 
     // This could be loosened up to make sure only the fastest changing dims
@@ -877,7 +878,7 @@ public:
   {
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
-    static_assert(RANK >= 2, "Only tensors of rank 2 and higher can be permuted.");
+    static_assert(RANK >= 1, "Only tensors of rank 1 and higher can be permuted.");
     cuda::std::array<shape_type, RANK> n;
     cuda::std::array<stride_type, RANK> s;
     [[maybe_unused]] bool done[RANK] = {0};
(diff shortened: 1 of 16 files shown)


include/matx/core/tensor.h
```diff
@@ -877,7 +878,7 @@ class tensor_t : public detail::tensor_impl_t<T,RANK,Desc> {
   {
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
 
-    static_assert(RANK >= 2, "Only tensors of rank 2 and higher can be permuted.");
+    static_assert(RANK >= 1, "Only tensors of rank 1 and higher can be permuted.");
```
cliffburdick, line 881: What's the reason to permute a 1D tensor?

include/matx/operators/pinv.h
```diff
@@ -0,0 +1,142 @@
...
+ * 
+ * @tparam OpA
+ *   Tensor or operator type of input A
+ * 
+ * @param a
+ *   Input tensor or operator
+ * @param rcond
+ *   Cutoff for small singular values. For stability, singular values
+ *   smaller than `rcond * largest_singular_value` are set to 0 for each matrix
+ *   in the batch. By default, `rcond` is approximately the machine epsilon of the tensor dtype.
```
cliffburdick, line 134: Does this need to specify what it returns?

include/matx/core/tensor.h
aayushg55: We have to take the max along the last dimension of the singular values vector, and max() calls permute. The vector could be of rank 1 if it isn’t batched.

include/matx/operators/pinv.h
aayushg55: Do you mean describing the shape of the calculated pinv tensor? Yes that would be good to add

cliffburdick (approved):
Very nice work

include/matx/operators/pinv.h
cliffburdick: I mean the function returns an operator, so should it have a @return that says that? But also since the operator is the user interface it should say something like "Returns a tensor of size mxn that represents the pseudo-inverse of the input"

aayushg55: Oh I see. I'll add an @return field

PR #743, opened 2024-08-30
Updated documentation for CPU support

aayushg55:
Deleted old `Limitations` page and added new `Executor Compatibility` page


docs_input/api/stats/corr/cov.rst
```diff
@@ -5,6 +5,10 @@ cov
 
 Compute a covariance matrix
 
+.. note::
+   This function is currently not supported with host-based executors (CPU)
```
cliffburdick, line 9: Looking at the `cov` code it seems like it should be supported on the host, but just hasn't been updated yet. I don't see any functions that would prevent it from working.

aayushg55: Yea, we just need to change the streams to executors

cliffburdick:
Overriding the failed CI since these are docs changes and the docs building passed.

PR #757, opened 2024-09-30
Create temp contiguous tensors if needed for sort

tbensonatl:
The sort implementation currently requires contiguous tensor view inputs. The run-time check for that requirement is currently disabled for release builds and thus those builds can produce incorrect results for non-contiguous tensors.

This change creates temporary contiguous allocations and copies the input when necessary to support sort(). This has the side effect of adding some operator input support for sort() because the operator will execute to populate the temporary tensor.

### DIFF
--- a/include/matx/transforms/cub.h
+++ b/include/matx/transforms/cub.h
@@ -563,7 +563,9 @@ inline void ExecSort(OutputTensor &a_out,
 
 #ifdef __CUDACC__
   static_assert(is_tensor_view_v<InputOperator>, "Sorting only accepts tensors for now (no operators)");
-  MATX_ASSERT_STR(a.IsContiguous(), matxInvalidType, "Tensor must be contiguous in memory for sorting");
+  if (! a.IsContiguous()) {
+    MATX_THROW(matxInvalidType, "Tensor must be contiguous in memory for sorting");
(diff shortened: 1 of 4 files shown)


include/matx/transforms/cub.h
```diff
@@ -563,7 +563,9 @@ inline void ExecSort(OutputTensor &a_out,
 
 #ifdef __CUDACC__
   static_assert(is_tensor_view_v<InputOperator>, "Sorting only accepts tensors for now (no operators)");
-  MATX_ASSERT_STR(a.IsContiguous(), matxInvalidType, "Tensor must be contiguous in memory for sorting");
+  if (! a.IsContiguous()) {
```
cliffburdick, line 566: We eventually should have checks that are for both debug and release

cliffburdick (approved):

include/matx/transforms/cub.h
tbensonatl: Agreed. In this case, I expect that we would currently never throw this exception because the caller of this function will allocate a contiguous tensor if necessary. I left it in place in case we rework things in the future to avoid the temporary allocation in some cases (e.g., by using CUB's merge sort).

PR #758, opened 2024-10-03
Fix regression in slice

cliffburdick:


luitjens (approved):
We should add a unit test to catch this bug.

PR #775, opened 2024-10-21
Remove relaxed constexpr

cliffburdick:


cmake/versions.json
```diff
@@ -1,10 +1,9 @@
 {
   "packages": {
     "CCCL": {
-      "version": "2.4.0",
+      "version": "2.7.0-rc2",
```
tbensonatl, line 4: Should we wait for the full 2.7 release to avoid needing to bump this again in the near future?

cliffburdick: My understanding is it's basically done, but they just haven't bumped that tag yet, but I will confirm

cliffburdick: They said it's not coming out until december, but that 2.7.0 was mostly done. The purpose of this PR is mostly to catch bugs; by turning relaxed-constexpr off we can't get those host/device bugs anymore, so I think it's probably worth it to just call this 2.7.0 and change if needed later.

tbensonatl: Agreed

tbensonatl (approved):

PR #783, opened 2024-10-27
Add example compiled with host compiler to catch regressions.

tmartin-gh:


cliffburdick:
Looks like it's getting error:

```
/home/jenkins/workspace/unit-tests/include/matx/transforms/cub.h: In member function 'T matx::detail::CustomArgMaxCmp::operator()(const T&, const T&) const':
/home/jenkins/workspace/unit-tests/include/matx/transforms/cub.h:1103:12: error: 'thrust' has not been declared
 1103 |     return thrust::get<1>(a) < thrust::get<1>(b) ? b : a;
      |            ^~~~~~
```

tmartin-gh:
Something is broken with CICD, it’s using my branch directly instead of merging first with main. The error reported shouldn’t be possible if the ifdef guard is present here https://github.com/NVIDIA/MatX/blob/834c31d91b805349dc31612fd41197de7dc39074/include/matx/transforms/cub.h#L1099 is

cliffburdick:
> Something is broken with CICD, it’s using my branch directly instead of merging first with main. The error reported shouldn’t be possible if the ifdef guard is present here
> 
> https://github.com/NVIDIA/MatX/blob/834c31d91b805349dc31612fd41197de7dc39074/include/matx/transforms/cub.h#L1099
> 
> is

It always uses your branch unless you've rebased on top of main

cliffburdick (approved):

PR #792, opened 2024-11-04
Modify concat op to enable concatenating float3.

nvjonwong:


examples/print_styles.cu
```diff
@@ -83,6 +99,32 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
   print(A4);
   // example-end print-example-1
 
+  // create stream
```
cliffburdick, line 102: I would leave out all this code since it's not really relevant to the fix, and we don't want to encourage people to use float3 types in tensors.

nvjonwong: Done.

cliffburdick (approved):

PR #794, opened 2024-11-05
Cudastf

sidelnik:
Initial updates to the build system to get Matx working with CUDASTF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,16 +119,18 @@ if (CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 11.5)
     message(FATAL_ERROR "MatX requires CUDA 11.5 or higher. Please update before using.")
 endif()
 
+set(CCCL_ENABLE_UNSTABLE ON)
 message(STATUS "Finding CCCL...")
 rapids_cpm_cccl(
     BUILD_EXPORT_SET matx-exports
     INSTALL_EXPORT_SET matx-exports
 )
 
-target_link_libraries(matx INTERFACE CCCL::CCCL)
+target_link_libraries(matx INTERFACE CCCL::CCCL CCCL::cudax)
+set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr --extended-lambda)
 
 # Set flags for compiling tests faster
-set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --threads 0 -ftemplate-backtrace-limit=0)
+set(MATX_CUDA_FLAGS ${MATX_CUDA_FLAGS} --threads 0 -ftemplate-backtrace-limit=0)
 
 # Hack because CMake doesn't have short circult evaluation
 if (NOT CMAKE_BUILD_TYPE OR "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ rapids_cpm_cccl(
 )
 
 target_link_libraries(matx INTERFACE CCCL::CCCL CCCL::cudax)
-set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr --extended-lambda)
+set(MATX_CUDA_FLAGS ${CMAKE_CUDA_FLAGS} --extended-lambda)
 
 # Set flags for compiling tests faster
 set(MATX_CUDA_FLAGS ${MATX_CUDA_FLAGS} --threads 0 -ftemplate-backtrace-limit=0)
--- a/cmake/versions.json
+++ b/cmake/versions.json
@@ -1,10 +1,9 @@
 {
   "packages": {
     "CCCL": {
-      "version": "2.7.0-rc2",
-      "git_shallow": true,
+      "version": "2.8.0",
       "git_url": "https://github.com/NVIDIA/cccl.git",
-      "git_tag": "10e915ac7b79a1ab3b9d7a795c621b47b122f513"
+      "git_tag": "cb1fce5e1cb7362940bd7e74ab8fbf01942b6264"
     }
   }
 }


examples/cgsolve.cu
```diff
@@ -83,6 +88,10 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
   (maxn = matx::max(sqrt(norm))).run(exec);
 
   exec.sync();
+#if 1
+  ctx.finalize();
```
cliffburdick, line 92: what is `finalize` used for vs `sync`? Could you hide the context in the executor so the user doesn't need it, and calling `exec.sync()` calls `finalize()`?

examples/fft_conv.cu
```diff
@@ -129,18 +138,30 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
     
   }
 
+#if 0
   cudaEventRecord(stop, stream);
```
cliffburdick, line 142: Eventually we should mask these events behind the executor as well so the timing is the same regardless of the executor.

include/matx/core/operator_utils.h
```diff
@@ -120,7 +120,7 @@ namespace matx {
     __MATX_HOST__ __MATX_INLINE__ void AllocateTempTensor(TensorType &tensor, Executor &&ex, ShapeType &&shape, typename TensorType::value_type **ptr) {
       const auto ttl_size = std::accumulate(shape.begin(), shape.end(), static_cast<index_t>(1),
                                   std::multiplies<index_t>()) * sizeof(typename TensorType::value_type);      
-      if constexpr (is_cuda_executor_v<Executor>) {
+      if constexpr (is_cuda_executor_v<Executor> || is_stf_executor_v<Executor>) {
```
cliffburdick, line 123: Should we make a separate `is_device_executor`?

include/matx/core/tensor.h
```diff
@@ -107,7 +108,7 @@ class tensor_t : public detail::tensor_impl_t<T,RANK,Desc> {
    * @param rhs Object to copy from
    */
   __MATX_HOST__ tensor_t(tensor_t const &rhs) noexcept
-      : detail::tensor_impl_t<T, RANK, Desc>{rhs.ldata_, rhs.desc_}, storage_(rhs.storage_)
+      : detail::tensor_impl_t<T, RANK, Desc>{rhs.ldata_, rhs.desc_, rhs.stf_ldata_}, storage_(rhs.storage_)
```
cliffburdick, line 111: It would be good to understand why this extra data member is needed, because this pointer exists on the device potentially many times, so it can increase the size of the operator.

examples/cgsolve.cu
caugonnet: finalize terminates everything in the stf context, it waits for asynchronous tasks, deletes internal resources etc...  you can only do it once, sync is more equivalent to a ctx.task_fence() which is a non blocking fence (it returns a CUDA stream, and waiting for that stream means everything was done).

I'd like to move finalize to the dtor of the executor, but there are some caveats if you define the executor as a static variable, is this allowed ? The caveat might be some inappropriate unload ordering of CUDA and STF libraries as usual ...

include/matx/core/operator_utils.h
caugonnet: Perhaps both stf and cuda executor needs to inherit from something common ? A trait class like is_device_executor is a good strategy to minimize the changes in the code to start with ...

include/matx/core/tensor.h
caugonnet: That's where a careful review of the design is needed ... Our logical data class tracks the use of a specific piece of data, your tensor seems to be a view to some data (with shapes and so on), so it's ok to have just the pointer and shapes, but in STF we do need to keep track of the internal state of the data (who owns a copy, which tasks depend on it, etc...). This is what the logical data does on your behalf and which your tensors cannot do by merely using the pointer.

One conservative take is to say that if you slice a tensor, this is the SAME logical data, so that further concurrent write accesses are serialized. This is sub-optimal when you have non overlapping slices but we cannot do better in a simple strategy. This ensures correctness but not optimality for concurrency

examples/cgsolve.cu
cliffburdick: Sounds good. I think the destructor is the right place. but does sync() work as expected?

caugonnet: @sidelnik is it doing a task fence with a stream sync ?

examples/fft_conv.cu
caugonnet: Yes this makes it look like the code is very different for both executors but that timing is the sole reason especially if finalize is moved to the dtor

include/matx/executors/stf.h
```diff
@@ -0,0 +1,253 @@
...
+      using matx_cuda = bool;  // signal this is a GPU executor
+      using matx_executor = bool; ///< Type trait indicating this is an executor
+//using cudastf::async_resources_handle;
+//using cudastf::stream_ctx;
+      /**
+       * @brief Construct a new stfExecutor with a stream
+       * 
+       * @param stream CUDA stream
+       */
+      stfExecutor(cudaStream_t stream) : stream_(stream) {
```
cliffburdick, line 63: What does a stream do here? I thought STF had its own internal streams?

include/matx/core/tensor.h
caugonnet: @cliffburdick you say it exists many times on the device, but isn't this a host only class ?

include/matx/core/tensor.h
```diff
@@ -177,6 +180,16 @@ class tensor_t : public detail::tensor_impl_t<T,RANK,Desc> {
     this->SetLocalData(storage_.data());
   }
 
+  template <typename S2 = Storage, typename D2 = Desc,
+            std::enable_if_t<is_matx_storage_v<typename remove_cvref<S2>::type> && is_matx_descriptor_v<typename remove_cvref<D2>::type>, bool> = true>
+  tensor_t(S2 &&s, D2 &&desc, T* ldata, std::optional<stf_logicaldata_type > *stf_ldata_) :
```
caugonnet, line 185: We need to do something about that type ... std::optional<stf_logicaldata_type > *stf_ldata_

The rationale is to be able to define a tensor before it is associated to an executor, so the logical data might be set lazily.

include/matx/core/tensor_impl.h
```diff
@@ -216,9 +229,18 @@ IGNORE_WARNING_PUSH_GCC("-Wmaybe-uninitialized")
...
     {
+      auto ldptr = new std::optional<stf_logicaldata_type>();
+      this->stf_ldata_ = ldptr;
     }
 IGNORE_WARNING_POP_GCC
 
+    template <typename DescriptorType, std::enable_if_t<is_matx_descriptor_v<typename remove_cvref<DescriptorType>::type>, bool> =     true>
+    __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ tensor_impl_t(T *const ldata,
+                    DescriptorType &&desc, std::optional<stf_logicaldata_type > *stf_ldata)
+        : ldata_(ldata), desc_{std::forward<DescriptorType>(desc)}, stf_ldata_(stf_ldata)
```
caugonnet, line 240: ::std::move(stf_ldata) ?

include/matx/core/tensor_impl.h
```diff
@@ -1080,9 +1106,54 @@ IGNORE_WARNING_POP_GCC
...
+    {
+      auto &ld = stf_ldata_->value();
+
+      using namespace cuda::experimental::stf;
+#if 0
+      data_place place = getDataPlace(Data());
+#endif
+
+      if (perm == 0) {
+          task.add_deps(ld.write());
```
caugonnet, line 1120: We could directly build a task_dep in CUDASTF, matching perm value with the type ... But it seems there is no such thing as a clean way to do this !

include/matx/core/tensor_impl.h
```diff
@@ -1080,9 +1106,54 @@ IGNORE_WARNING_POP_GCC
...
+            /* Don't create a new logical data for a tensor if it alread had one created previously */
+            if (stf_ldata_ && stf_ldata_->has_value()) { return; }
+
+            auto ctx = ex.getCtx();
+#if 0
+            // Determine the type of memory that was allocated ie. host/managed/etc
+            place = getDataPlace(Data());
+#endif
+
+            *stf_ldata_ = ctx.logical_data(cuda::experimental::stf::void_interface());
```
caugonnet, line 1153: Some comment would be welcome here :)

include/matx/core/utils.h
```diff
@@ -45,6 +45,30 @@
 
 namespace matx {
 namespace detail {
+
+#if 0
+__MATX_INLINE__ cuda::experimental::stf::data_place getDataPlace(void *ptr) {
```
caugonnet, line 50: Why don't we keep it ? Note that for void data interface it's not super critical but still ...

include/matx/core/utils.h
```diff
@@ -45,6 +45,30 @@
...
+                        return data_place::managed;
+                    case MATX_HOST_MEMORY:
+                    case MATX_HOST_MALLOC_MEMORY:
+                        return data_place::host;
+                    case MATX_DEVICE_MEMORY:
+                    case MATX_ASYNC_DEVICE_MEMORY:
+                        return data_place::current_device();
+                    case MATX_INVALID_MEMORY:
+                        //std::cout << "Data kind is invalid: assuming managed memory\n";
+                        return data_place::managed;
```
caugonnet, line 64: this seems like an error

include/matx/executors/stf.h
```diff
@@ -0,0 +1,253 @@
...
+                    op.apply_dep_to_task(tsk); // recursively find the tensors from the tree to apply deps
+                    //std::cout << "Start launch task. Rank = " << Op::Rank() << " " << op.str() << '\n';
+                    tsk->*[&](cudaStream_t s) { 
+                        detail::matxOpT0Kernel<<<blocks, threads, 0, s>>>(op);
+                    };
+                    //std::cout << "End launch task.\n";
+            }
+            else {
+                //std::cout << " RANK 0 not on LHS operator = " << op.str() << '\n';
+                detail::matxOpT0Kernel<<<blocks, threads, 0, stream_>>>(op);
```
caugonnet, line 133: Why do we sometimes use something without a task ?

include/matx/executors/stf.h
```diff
@@ -0,0 +1,253 @@
...
+          }
+          else {
+            cuda::std::array<index_t, Op::Rank()> sizes;
+            for (int i = 0; i < Op::Rank(); i++) {
+              sizes[i] = op.Size(i);
+            }        
+
+            bool stride = detail::get_grid_dims<Op::Rank()>(blocks, threads, sizes, 256);
+
+            if constexpr (Op::Rank() == 1) {
```
caugonnet, line 144: It looks like we could factorize all that constexpr cascade, and move the constexpr tests into the lambda ?

include/matx/generators/generator1d.h
```diff
@@ -54,6 +54,9 @@ namespace matx
             return f_(pp_get<Dim>(indices...));
           }
 
+        template <typename Task>
+        __MATX_INLINE__ void apply_dep_to_task([[maybe_unused]] Task &&task, [[maybe_unused]] int perm=1) const noexcept { }
```
caugonnet, line 58: So this operator is defined per operator, and is STF specific ? it's not part of the executor nor relying on overloads / traits ?

include/matx/operators/conv.h
```diff
@@ -149,16 +149,45 @@ namespace matx
...
+              tsk.set_symbol("conv_task");
+
+              auto output = cuda::std::get<0>(out);
+              output.PreRun(out_dims_, std::forward<Executor>(ex));
+              output.apply_dep_to_task(tsk, 0);
+              a_.apply_dep_to_task(tsk, 1);
+              b_.apply_dep_to_task(tsk, 1);
+
+              tsk->*[&](cudaStream_t s) {
+                  auto exec = cudaExecutor(s);
```
caugonnet, line 175: So create a nested MatX executor, is that legal ?

include/matx/core/tensor_impl.h
```diff
@@ -110,13 +113,15 @@ class tensor_impl_t {
...
       swap(lhs.desc_, rhs.desc_);
+      swap(lhs.stf_ldata_, rhs.stf_ldata_);
     }
 
     /**
      * Constructor for a rank-0 tensor (scalar).
      */
     tensor_impl_t() {
-
+      auto ldptr = new std::optional<stf_logicaldata_type>();
```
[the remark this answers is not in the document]
cliffburdick: This won't compile anymore since we don't allow std:: types on the device. It might work with cuda::std::optional, but we don't use that anywhere currently.

include/matx/core/tensor_impl.h
```diff
@@ -1080,9 +1106,54 @@ IGNORE_WARNING_POP_GCC
...
     template <typename ShapeType, typename Executor>
     __MATX_INLINE__ void PreRun([[maybe_unused]] ShapeType &&shape, [[maybe_unused]] Executor &&ex) const noexcept
     {
+        using namespace cuda::experimental::stf;
+        data_place place;
+
+        if constexpr (is_cuda_executor_v<Executor>) {
+            return;
+        }
+        else if constexpr (!is_cuda_executor_v<Executor>) {
```
cliffburdick, line 1142: just `else`?

include/matx/core/tensor_impl.h
```diff
@@ -1094,6 +1165,9 @@ IGNORE_WARNING_POP_GCC
   protected:
     T *ldata_;
     Desc desc_;
+
+  public:
+    mutable std::optional<stf_logicaldata_type > *stf_ldata_;
```
cliffburdick, line 1170: As discussed before this won't work since we can't use std:: objects on the device. It might work with `cuda::std::optional`, but we'd likely need to justify the overhead vs other options

include/matx/operators/constval.h
```diff
@@ -55,6 +55,9 @@ namespace matx
         __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ T operator()(Is...) const { 
           return v_; };
 
+      template <typename Task>
+      __MATX_INLINE__ void apply_dep_to_task([[maybe_unused]] Task &&task, [[maybe_unused]] int perm) const noexcept { }
```
cliffburdick, line 59: Operator members typically use camel case format

include/matx/operators/all.h
```diff
@@ -72,9 +72,35 @@ namespace detail {
...
-        all_impl(cuda::std::get<0>(out), a_, ex);
+        auto output = cuda::std::get<0>(out);
+        // stfexecutor case
+        if constexpr (!is_cuda_executor_v<Executor>) {
+            auto ctx = ex.getCtx();
+            auto tsk = ctx.task();
+            tsk.set_symbol("all_task");
+
+            output.PreRun(out_dims_, std::forward<Executor>(ex));
+            output.apply_dep_to_task(tsk, 0);
```
cliffburdick, line 92: Why isn't `apply_dep_to_task` just part of `PreRun`? It looks like it's called in the same place

include/matx/operators/conv.h
cliffburdick: I think it should be fine. The cache is ultimately what would possibly have side effects

include/matx/operators/fft.h
```diff
@@ -145,24 +145,68 @@ namespace matx
...
-            }
-            else {
-              ifft_impl(cuda::std::get<0>(out), a_, fft_size_, norm_, ex);
-            }
-          }
-          else {
-            if constexpr (std::is_same_v<FFTType, fft_t>) { 
-              fft_impl(permute(cuda::std::get<0>(out), perm_), permute(a_, perm_), fft_size_, norm_, ex);
+            // stfexecutor case
+            if constexpr (!is_cuda_executor_v<Executor>) {
```
cliffburdick, line 159: Do you want this to run for the host executor too?

include/matx/operators/fft.h
```diff
@@ -145,24 +145,68 @@ namespace matx
...
+                auto ctx = ex.getCtx();
+                auto tsk = ctx.task();
+                tsk.set_symbol("fft_task_no_perm");
+
+                auto output = cuda::std::get<0>(out);
+                output.PreRun(out_dims_, std::forward<Executor>(ex));
+                output.apply_dep_to_task(tsk, 0);
+                a_.apply_dep_to_task(tsk, 1);
+
+                tsk->*[&](cudaStream_t s) {
```
cliffburdick, line 169: Rather than checking if this is not a cuda executor, then creating one inside, can it somehow pull a stream from STF and just use that here?

examples/cgsolve.cu
sidelnik: @caugonnet , sync() should be calling ctx.task_fence() now. I agree, I think we should place the ctx.finalize() inside the stf executor dtor

include/matx/executors/stf.h
sidelnik: @cliffburdick In STF you can create nested/localized contexts & streams from existing (non-STF created) streams. This allows STF mechanisms to be correctly synchronized within the existing stream ecosystem. @caugonnet correct me if I am wrong.

include/matx/core/tensor_impl.h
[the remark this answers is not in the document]
caugonnet: @sidelnik now that we have the notion of logical_token, i believe we might simplify that. Maybe rename stf_logicaldata_type to stf_token ?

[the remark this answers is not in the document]
caugonnet: The risk with token is that if we get it wrong, it's easier to mess things : with a logical data, until you do have some "value" in it, you can't read it and you'll get runtime errors. If you have aliases which would currently use the same token under the hood, it would also have to use the same token when creating the aliased data.

include/matx/transforms/cgsolve.h
```diff
@@ -120,15 +120,19 @@ namespace matx
       auto pApc = clone<VRANK>(pAp, clone_shape);
     
       // A*X
-      (Ap = matvec(A, X)).run(stream);
+      //(Ap = matvec(A, X)).run(stream);
+      (Ap = matvec(A, X)).run(exec);
```
caugonnet, line 124: is that the same to call run(exec) and run(stream) when we have a "classic" executor ? (won't it trigger much more work ?)

include/matx/operators/conv.h
caugonnet: What happens exactly in the dtor of the executor @cliffburdick, nothing special like a stream sync ?

include/matx/core/tensor.h
cliffburdick: tensor_t is host/device, but tensor_impl_t is device-only

include/matx/operators/conv.h
cliffburdick: No, it doesn't do anything

include/matx/core/tensor.h
caugonnet: Then i'm even surprised a logical_data<T> can exist in device code, or the storage for it ! But this may be a pointer to an optional logical data ... We need to improve that

cliffburdick: ugh I mistyped. tensor_t is ONLY on the host. tensor_impl_t is both.

caugonnet: Still, it's surprising that we allow the logical data pointer to go on a device

cliffburdick: ldata is local data, and is ultimately just a raw pointer that points to the data needed on the device. This may be the same as the base pointer, or it may be something like a strided/offset pointer.

PR #818, opened 2025-01-08
Add cross product operator

mfzmullen:
Takes in two operators of size A0 x ... x An x 3 (or 2) and size B0 x ... x Bn x (3 or 2) with matching batch sizes and computes the cross product. If both input operators have size = 2 on the last dimension, the size of the output's last dimension is 1 and only contains the z component of the output. This is contrary, but as similar to, the NumPy implementation which drops the rank of the output by 1 if both inputs are in-plane vectors. Doing so in MatX does not seem straightforward since `Rank` is a static method and we may not know the sizes of A and B (e.g. 2 or 3 on the last dimension) at the time the rank is determined, but those sizes affect the rank of `cross` if we followed NumPy's approach.

### DIFF
--- a/include/matx/operators/cross.h
+++ b/include/matx/operators/cross.h
@@ -189,8 +189,8 @@ namespace matx
    *  
    * @tparam OpA Type of input tensor 1
    * @tparam OpB Type of input tensor 2
-   * @param OpA Input tensor 1
-   * @param OpB Input tensor 2
+   * @param A Input tensor 1
+   * @param B Input tensor 2
    * @return cross operator 
    */
   template <typename OpA, typename OpB>
--- a/include/matx/operators/cross.h
+++ b/include/matx/operators/cross.h
@@ -74,9 +74,7 @@ namespace matx
           MATX_ASSERT_STR(a_.Size(a_.Rank() - 1) == 3 || a_.Size(a_.Rank() - 1) == 2, matxInvalidSize, "Last dimension of A must have size 2 or 3.")
           MATX_ASSERT_STR(b_.Size(b_.Rank() - 1) == 3 || b_.Size(b_.Rank() - 1) == 2, matxInvalidSize, "Last dimension of B must have size 2 or 3.")
         
-          //all input vector sizes (3 x 3, 2 x 3, 3 x 2, 2 x 2) will output to a length-3 vector (plus batch dims)
-          out_dims_[0] = 3;
-          for (int32_t i = 1; i < out_rank; i++) {
+          for (int32_t i = 0; i < out_rank - 1; i++) {
             if (i < a_.Rank()){
               out_dims_[i] = a_.Size(i);
             }
@@ -84,9 +82,12 @@ namespace matx
               out_dims_[i] = b_.Size(i);
             }
           }
-        
-          //promote output to 3D
+          
+          //mimic NumPy cross as closely as possible
           if(isA2D_ && isB2D_){
+            out_dims_[out_dims_.size() - 1] = 1;
+          }
+          else{
             out_dims_[out_dims_.size() - 1] = 3;
           }
         };
@@ -114,7 +115,7 @@ namespace matx
 
           //lots of if-elses, but similar to numpy implementation
           
-          if (idxOut == 2){
+          if (idxOut == 2 || (isA2D_ && isB2D_)){
             return a0 * b1 - a1 * b0;
           }
 
@@ -136,16 +137,13 @@ namespace matx
             //idxOut == 1
             return -a0 * b2;
           }
-          else if(!isA2D_ && isB2D_){
+          else{// !isA2D_ && isB2D_, case of both 2D are covered in the first if statement
             auto a2 = get_value(a_, idx2);
             if (idxOut == 0){
                 return -a2 * b1;
             }
             //idxOut == 1
             return a2 * b0;
-          }
-          else{ //both are 2D, just need to return 0 with correct type and shape
-            return static_cast<value_type>(0)*a0;
           }
           
         }
--- a/include/matx/operators/cross.h
+++ b/include/matx/operators/cross.h
@@ -1,0 +1,202 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// COpBright (c) 2021, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above cOpBright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above cOpBright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the cOpBright holder nor the names of its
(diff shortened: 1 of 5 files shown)


include/matx/operators/cross.h
```diff
@@ -0,0 +1,202 @@
...
+        };
+
+        template <typename... Is>
+        __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) operator()(Is... indices) const
+        {
+          cuda::std::array idx{indices...};
+          auto idxOut = idx[idx.size() - 1];
+
+          //create references to individual slices for ease of notation
+          cuda::std::array idx0(idx);
```
cliffburdick, line 101: I think this is creating an unnecessary copy in the constructor. Instead these can be created by just using the type and size of `idx`.

include/matx/operators/cross.h
```diff
@@ -0,0 +1,202 @@
...
+          cuda::std::array idx1(idx);
+          cuda::std::array idx2(idx);
+
+          idx0[idx0.size() - 1] = 0LL;
```
cliffburdick, line 105: Aren't the rest of the array elements uninitialized here?

include/matx/operators/cross.h
```diff
@@ -0,0 +1,202 @@
...
+          else if(!isA2D_ && isB2D_){
+            auto a2 = get_value(a_, idx2);
+            if (idxOut == 0){
+                return -a2 * b1;
+            }
+            //idxOut == 1
+            return a2 * b0;
+          }
+          else{ //both are 2D, just need to return 0 with correct type and shape
+            return static_cast<value_type>(0)*a0;
```
cliffburdick, line 148: Why does it need to multiple 0 by another number instead of returning 0?

test/00_operators/OperatorTests.cu
```diff
@@ -2012,6 +2012,105 @@ TYPED_TEST(OperatorTestsFloatAllExecs, Toeplitz)
...
+    a.SetVals({{1, 0, 0}, 
+              {0, 0.9, 0}, 
+              {0, 0, 0.8}});
+    b.SetVals({{0, 0, 0.7}, 
+              {0.6, 0, 0}, 
+              {0, 0.5, 0}});
+    auto expect = make_tensor<TestType>({3, 3});
+    for (int i = 0; i < 3; i++) {
+      for (int j = 0; j < 3; j++) {
+          expect(i,j) = a(i, (j+1)%3) * b(i, (j+2)%3) - a(i, (j+2)%3) * b(i, (j+1)%3);
```
cliffburdick, line 2035: Can you use numpy's cross function here rather than your own implementation? We try to do that as much as possible. It's pretty simple to do -- you can look at examples of `fft` to see how.

cliffburdick (changes requested):
Looks great! Thanks for doing this

include/matx/operators/cross.h
mfzmullen: Good question, I don't know. I think this was a relic of something I tried early on. In any case, this line was removed when updating to use numpy for unit tests, since `matx::cross` now outputs with size 1 on the last dim if both inputs are 2D, as that made it more functionally similar to numpy.

test/00_operators/OperatorTests.cu
mfzmullen: also, the tests with half types were failing sometimes, and generally the difference was just above the cutoff threshold for pass/fail. Instead of changing the threshold or making it type-dependent, I just removed the half types from the unit tests for cross.

cliffburdick: Hi @mfzmullen can you leave the half tests in and make the threshold larger? something like this:

https://github.com/NVIDIA/MatX/blob/main/test/00_transform/FFT.cu#L65

mfzmullen: Thanks for this while I collect my wits on contributing. Updated to change threshold for non-complex half types.

cliffburdick (approved):

cliffburdick:
Hi @mfzmullen, I apologize I didn't see this earlier, but the documentation needs to be updated:

```
/home/jenkins/workspace/unit-tests/include/matx/operators/cross.h:188: error: argument 'OpA' of command @param is not found in the argument list of matx::cross(const OpA &A, const OpB &B) (warning treated as error, aborting now)
```

mfzmullen:
@cliffburdick no worries, my bad for missing that in the first place! Just pushed

cliffburdick:
Hi @mfzmullen now It's getting:

```
/home/jenkins/workspace/unit-tests/docs_input/api/linalg/other/cross.rst:16: WARNING: Include file '/home/jenkins/workspace/unit-tests/test/00_Operator/OperatorTests.cu' not found or reading it failed [docutils]`
```

You can test this yourself by "make sphinx" if you want it to run in a local environment.

mfzmullen:
@cliffburdick got the documentation built locally with latest push. Again my apologies on the mess with that, but thank you for your patience.

cliffburdick:
Merged. Thanks @mfzmullen !

cliffburdick:
@mfzmullen it looks like the unit test is failing: 
```
[  FAILED  ] OperatorTestsFloatNonComplexAllExecs/4.Cross, where TypeParam = cuda::std::__4::tuple<matx::matxHalf<__nv_bfloat16>, matx::cudaExecutor>
```

mfzmullen:
@cliffburdick strange, I played with the tolerance a bit for the half types and wasn't getting any failures with the current setting after a few dozen runs. I can look at it a bit more.

cliffburdick:
Here is the error:
```
Comparison failed at /home/jenkins/workspace/unit-tests/test/00_operators/OperatorTests.cu:2035:0/1: val=4.4375 file=4.4078 (out)
/home/jenkins/workspace/unit-tests/test/00_operators/OperatorTests.cu:2035: Failure
Failed
[  FAILED  ] OperatorTestsFloatNonComplexAllExecs/4.Cross, where TypeParam = cuda::std::__4::tuple<matx::matxHalf<__nv_bfloat16>, matx::cudaExecutor> (0 ms)
```

mfzmullen:
@cliffburdick if I change the threshold to 0.06 for the half types and run
`./test/matx_test --gtest_filter="*Cross*" --gtest_repeat=1000 --gtest_break_on_failure`

I had it pass twice and failed once after 623 iterations. What failure probability is acceptable?

mfzmullen:
While there isn't an uncertainty, each repetition uses different randomly generated inputs, so the difference between the numpy and matx versions will change on each repetition. I'll push a change soon.

mfzmullen:
Oh got it. I see that in some of the generators now, but not in the ones I used as examples (toeplitz). I can add that to my PR.

cliffburdick:
It looks like it's still failing with a larger threshold. I will make it even larger since it appears to be just a precision issue:
```
Comparison failed at /home/jenkins/workspace/unit-tests/test/00_operators/OperatorTests.cu:2035:2/2: val=-4 file=-4.02057 (out)
/home/jenkins/workspace/unit-tests/test/00_operators/OperatorTests.cu:2035: Failure
```

mfzmullen:
That difference is already less than thresh (=0.08). That is with a double or float perhaps? All these tests passed on my local machine.

The tests need to pass before merging to main or no? Just curious why they seem to be failing sometimes even with the consistent seed in the rng now.

cliffburdick:
It looks like it's working now. Thanks!

PR #821, opened 2025-01-14
introduce a versatile sparse tensor type to MatX (experimental)

aartbik:
This PR introduces the implementation of a single versatile sparse tensor type that uses a tensor format DSL (Domain Specific Language) to describe a vast space of storage formats. Although the tensor format can easily define many common storage formats (such as Dense, COO, CSR, CSC, BSR), it can also define many less common storage formats. In addition, the tensor format DSL can be extended to include even more storage formats in the future.

This first PR simply introduces all storage details for the single versatile sparse tensor type, together with some factory methods for constructing COO, CSR, and CSC sparse matrices from MatX buffers. Later PRs will introduce more general ways of constructing sparse tensors (e.g. from file) and actual operations like SpMV and SpMM using cuSPARSE.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -11,6 +11,7 @@ set(examples
     mvdr_beamformer
     pwelch
     resample_poly_bench
+    sparse_tensor
     spectrogram
     spectrogram_graph
     spherical_harmonics
--- a/include/matx/core/make_sparse_tensor.h
+++ b/include/matx/core/make_sparse_tensor.h
@@ -1,0 +1,104 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2025, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its
+//    contributors may be used to endorse or promote products derived from
+//    this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+/////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include "matx/core/sparse_tensor.h"
+
+namespace matx {
+namespace experimental {
+
+//
+// MatX uses a single versatile sparse tensor type that uses a tensor format
+// DSL (Domain Specific Language) to describe a vast space of storage formats.
+// This file provides a number of convenience factory methods that construct
+// sparse tensors in well-known storage formats, like COO, CSR, and CSC,
+// directly from the constituent buffers. More factory methods can easily be
+// added as the need arises.
+//
+
+// Constructs a sparse matrix in COO format directly from the values and
+// the two coordinates vectors. The entries should be sorted by row, then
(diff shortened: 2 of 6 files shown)


examples/CMakeLists.txt
```diff
@@ -67,4 +68,4 @@ target_link_libraries(test_host matx::matx)
 
 if(MATX_EN_PYBIND11)
     add_subdirectory(python_integration_sample)
-endif()
\ No newline at end of file
+endif()
```
cliffburdick, line 71: Seems like a mistake?

aartbik: Code formatter added a newline at end of file, let me try to minimize changes....

aartbik: I split this (and a few other minor) changes out into their own PR, just so that this PR is purely focused on the new sparse functionality, without any distractions....

include/matx/core/make_sparse_tensor.h
```diff
@@ -0,0 +1,139 @@
...
+// This file provides a number of convenience factory methods that construct
+// sparse matrices in well-known storage formats, like COO, CSR, and CSC,
+// directly from the constituent buffers. More factory methods can easily be
+// added as the need arises.
+//
+
+// Internal helper method for sparse matrices (2-dim tensors).
+template <typename VAL, typename CRD, typename POS, typename TF>
+static __MATX_INLINE__ auto
+make_sparse_matrix(const index_t (&shape)[2], VAL *val, size_t vsz, CRD *c0,
```
cliffburdick, line 52: Probably needs a doxygen comment explaining each parameter

include/matx/core/make_sparse_tensor.h
```diff
@@ -0,0 +1,139 @@
...
+                         decltype(sp0)>(shape, std::move(sv),
+                                        {std::move(sc0), std::move(sc1)},
+                                        {std::move(sp0), std::move(sp1)});
+}
+
+// Constructs a sparse matrix in COO format directly from the values and
+// coordinates arrays. The entries should be sorted by row, then column.
+// Duplicate entries should not occur. Explicit zeros may be stored.
+template <typename VAL, typename CRD>
+auto make_coo(VAL *val, CRD *row, CRD *col, const index_t (&shape)[2],
```
cliffburdick, line 79: trying not to be too verbose, but should this be called `make_tensor_coo` to fit with the dense naming? or maybe `make_tensor_sp_coo`?

include/matx/core/make_sparse_tensor.h
```diff
@@ -0,0 +1,139 @@
...
+              size_t nse, bool owning = false) {
+  const size_t vsz = nse * sizeof(VAL);
+  const size_t csz = nse * sizeof(CRD);
+  const size_t psz = (shape[0] + 1) * sizeof(POS);
+  return make_sparse_matrix<VAL, CRD, POS, CSR>(
+      shape, val, vsz, nullptr, 0, col, csz, nullptr, 0, rowp, psz, owning);
+}
+
+// As above for CSR, operating directly on MatX 1-dim tensors for the buffers.
+template <typename ValTensor, typename PosTensor, typename CrdTensor>
```
cliffburdick, line 110: How does this one ensure that the data stays there and doesn't go out of scope if it uses `Data()` directly?

include/matx/core/print.h
```diff
@@ -149,7 +149,25 @@ namespace matx {
           fprintf(fp, ", ");
       }
 
-      if constexpr (is_tensor_view_v<Op>)
+      if constexpr (is_sparse_tensor_v<Op>)
```
cliffburdick, line 152: Should this also change line 143 to either display Sparse Tensor, Dense Tensor, or Operator?

include/matx/core/print.h
```diff
@@ -543,7 +561,40 @@ namespace matx {
...
       cudaDeviceSynchronize();
-      if constexpr (is_tensor_view_v<Op>) {
+      if constexpr (is_sparse_tensor_v<Op>) {
+        using Format = typename Op::Format;
+	index_t nse = op.Nse();
+        fprintf(fp, "nse    = %" MATX_INDEX_T_FMT "\n", nse);
+        fprintf(fp, "format = ");
+	Format::print();
+	//
+        // TODO: how to access pos/crd/val data properly in MatX style?
```
cliffburdick, line 571: Is POSData a matx tensor? Then operator() is correct

cliffburdick (commented):
Looks good!

We do need some unit tests and documentation to go along with it so people can start using it. I think the sparse unit tests should have its own directory inside of tests/ for now.

include/matx/core/make_sparse_tensor.h
aartbik: Added a doxygen comment to the helper method

include/matx/core/make_sparse_tensor.h
aartbik: changed all method names accordingly

include/matx/core/make_sparse_tensor.h
aartbik: That clearly indicates a lack of understanding from my part. How can I ensure that ownership and ref counts are properly dealt with when using MatX vectors to provide the 1-dim buffer contents?

include/matx/core/print.h
aartbik: Good idea. Done.

include/matx/core/print.h
aartbik: Well, technically it is the .data() field of a Storage container, possibly taken from a matx vector (but one could also allocate their own buffer directly). Is () op still correct then?

cliffburdick: I think in general we should require a tensor to be passed in for those values and not a storage type. I suppose storage can be used if it's always 1D linear memory, but the tensor is more like the other APIs and would give us flexibility later.

include/matx/core/make_sparse_tensor.h
cliffburdick: If you copy a storage buffer the reference count is handled properly. But if you just access .Data() directly from the tensor there is no guarantee that the tensor will stay in scope when it's used. If you want to keep the storage types you can simply copy them via the make_csr parameters instead of raw pointers. This would increment the reference count inside of the storage class (if there is one)

aartbik: Okay, let me redesign the make sparse tensor methods a bit then to ensure that is the case

include/matx/core/print.h
aartbik: see above, we will need to only provide factory methods that take matx tensors

aartbik (commented):
Thanks for your excellent codereview, Cliff.
I learned a lot, and the code looks much better now!
Thanks!

cliffburdick (approved):

PR #823, opened 2025-01-15
add a to-string method for memory space

aartbik:


include/matx/core/allocator.h
```diff
@@ -413,6 +413,15 @@ struct matx_allocator {
   }  
 };
 
-
+inline std::string SpaceString(matxMemorySpace_t space) {
```
cliffburdick, line 416: This should be `__MATX_INLINE__`

cliffburdick (changes requested):
Just a small change

include/matx/core/allocator.h
aartbik: Changed (but please note that most methods in this file still use "inline", should that change too?)

cliffburdick: Yes, we try to use force inline in most places, so please change to the macro instead

aartbik: done!

cliffburdick (approved):

PR #826, opened 2025-01-16
refactored permutation (less code dup)

aartbik:
this change avoids some of the code duplication
within the permute / no-permute branches by
refactoring output permutation into prefix

### DIFF
--- a/include/matx/operators/conv.h
+++ b/include/matx/operators/conv.h
@@ -154,12 +154,11 @@ namespace matx
           MATX_ASSERT_STR(!(is_host_executor_v<Executor> && method_ == MATX_C_METHOD_DIRECT), matxNotSupported, "direct conv1d() only supports the CUDA executor currently");
           MATX_STATIC_ASSERT_STR((Rank() == cuda::std::tuple_element_t<0, remove_cvref_t<Out>>::Rank()), 
                 matxInvalidParameter, "conv1d: inputs and outputs must have same rank to use conv1d with axis parameter");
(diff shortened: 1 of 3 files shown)


include/matx/operators/conv.h
```diff
@@ -154,12 +154,11 @@ namespace matx
           MATX_ASSERT_STR(!(is_host_executor_v<Executor> && method_ == MATX_C_METHOD_DIRECT), matxNotSupported, "direct conv1d() only supports the CUDA executor currently");
           MATX_STATIC_ASSERT_STR((Rank() == cuda::std::tuple_element_t<0, remove_cvref_t<Out>>::Rank()), 
                 matxInvalidParameter, "conv1d: inputs and outputs must have same rank to use conv1d with axis parameter");
+          auto outop = cuda::std::get<0>(out);
           if constexpr (!std::is_same_v<PermDims, no_permute_t>) {
-            conv1d_impl(permute(cuda::std::get<0>(out), perm_), a_, b_, mode_, method_, ex);
-          }
-          else {
-            conv1d_impl(cuda::std::get<0>(out), a_, b_, mode_, method_, ex);
+            outop = permute(outop, perm_);
```
cliffburdick, line 159: I'm surprised this compiles at all. The operator returned above is not the same type as what `permute` returns. Do you know what's happening?

aartbik: If the type matters for the dispatch, then we should probably just keep the branchy code....

PR #836, opened 2025-01-22
Build a pos[0] = (0,nse) structure for COO

aartbik:
Rationale:

The sparse COO interface typically used by SciPy and PyTorch does not use the positions buffers, but only the coordinate buffers. However, the versatile sparse tensor DSL requires a positions indicator for each compressed level. This change adds this automatically, with no changes to the user facing API.

After this, the "locate" operation will be much more regular for all level format: dense, compressed, singleton.

### DIFF
--- a/include/matx/core/make_sparse_tensor.h
+++ b/include/matx/core/make_sparse_tensor.h
@@ -55,11 +55,23 @@ auto make_tensor_coo(ValTensor &val, CrdTensor &row, CrdTensor &col,
   static_assert(ValTensor::Rank() == 1 && CrdTensor::Rank() == 1);
   using VAL = typename ValTensor::value_type;
   using CRD = typename CrdTensor::value_type;
-  using POS = int; // no positions, although some forms use [0,nse]
+  using POS = index_t;
+  // Note that the COO API typically does not involve positions.
+  // However, under the formal DSL specifications, the top level
+  // compression should set up pos[0] = {0, nse}. This is done
+  // here, using the same memory space as the other data.
+  POS *ptr;
+  matxMemorySpace_t space = GetPointerKind(val.GetStorage().data());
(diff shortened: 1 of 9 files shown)


include/matx/core/make_sparse_tensor.h
```diff
@@ -55,11 +55,23 @@ auto make_tensor_coo(ValTensor &val, CrdTensor &row, CrdTensor &col,
...
   using VAL = typename ValTensor::value_type;
   using CRD = typename CrdTensor::value_type;
-  using POS = int; // no positions, although some forms use [0,nse]
+  using POS = index_t;
+  // Note that the COO API typically does not involve positions.
+  // However, under the formal DSL specifications, the top level
+  // compression should set up pos[0] = {0, nse}. This is done
+  // here, using the same memory space as the other data.
+  POS *ptr;
+  matxMemorySpace_t space = GetPointerKind(val.GetStorage().data());
```
cliffburdick, line 64: Do you need to check for errors here?

aartbik: do you mean the consistency of memory spaces? that is done inside the sparse tensor constructor itself, right? or some other sort of errors?

cliffburdick: I mean what if the space is not found or is an error like `MATX_INVALID_MEMORY`

aartbik: Ah, I could intercept, but the allocator already has a MATX_THROW when the space is invalid right after that. Can we simply rely on that?

aartbik:
Ah, I see the follow-up PR already went it, effectively making this a zero-diff PR (this PR was supposed to make the diff of the getter() PR less ;-)

So, this one can close

PR #837, opened 2025-01-22
Implement the () operator on sparse tensors

aartbik:
Note that is a fully functional "locator" for the () operator that works for *all* versatile sparse tensors. Currently, the operator is defined at sparse_tensor level, but it should be moved into tensor_impl after this. Also, clients should always be aware that the () operator for compressed levels are *not* random-access, but involve a search to find if the element is stored.

### DIFF
--- a/include/matx/core/make_sparse_tensor.h
+++ b/include/matx/core/make_sparse_tensor.h
@@ -55,11 +55,23 @@ auto make_tensor_coo(ValTensor &val, CrdTensor &row, CrdTensor &col,
   static_assert(ValTensor::Rank() == 1 && CrdTensor::Rank() == 1);
   using VAL = typename ValTensor::value_type;
   using CRD = typename CrdTensor::value_type;
-  using POS = int; // no positions, although some forms use [0,nse]
+  using POS = index_t;
+  // Note that the COO API typically does not involve positions.
+  // However, under the formal DSL specifications, the top level
+  // compression should set up pos[0] = {0, nse}. This is done
+  // here, using the same memory space as the other data.
+  POS *ptr;
+  matxMemorySpace_t space = GetPointerKind(val.GetStorage().data());
+  matxAlloc((void **)&ptr, 2 * sizeof(POS), space, 0);
+  ptr[0] = 0;
+  ptr[1] = val.Size(0);
+  raw_pointer_buffer<POS, matx_allocator<POS>> topp{ptr, 2 * sizeof(POS),
+                                                    owning};
+  basic_storage<decltype(topp)> tp{std::move(topp)};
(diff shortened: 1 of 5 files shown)


include/matx/core/make_sparse_tensor.h
```diff
@@ -55,11 +55,23 @@ auto make_tensor_coo(ValTensor &val, CrdTensor &row, CrdTensor &col,
...
   using VAL = typename ValTensor::value_type;
   using CRD = typename CrdTensor::value_type;
-  using POS = int; // no positions, although some forms use [0,nse]
+  using POS = index_t;
+  // Note that the COO API typically does not involve positions.
+  // However, under the formal DSL specifications, the top level
+  // compression should set up pos[0] = {0, nse}. This is done
+  // here, using the same memory space as the other data.
+  POS *ptr;
+  matxMemorySpace_t space = GetPointerKind(val.GetStorage().data());
```
cliffburdick, line 64: Same comment to check for an error type

include/matx/core/make_sparse_tensor.h
```diff
@@ -55,11 +55,23 @@ auto make_tensor_coo(ValTensor &val, CrdTensor &row, CrdTensor &col,
...
+  matxAlloc((void **)&ptr, 2 * sizeof(POS), space, 0);
```
cliffburdick, line 65: We should pass a stream here if this is a stream-oriented allocation, but this is fine for now.

include/matx/core/sparse_tensor_format.h
```diff
@@ -255,42 +255,48 @@ template <int D, typename... LvlSpecs> class SparseTensorFormat {
...
-      lvls[L] = dims[ftype::expr::di];
-    } else if constexpr (ftype::expr::op == LvlOp::Div) {
-      lvls[L] = dims[ftype::expr::di] / ftype::expr::cj;
-    } else if constexpr (ftype::expr::op == LvlOp::Mod) {
-      lvls[L] =
-          asSize ? ftype::expr::cj : (dims[ftype::expr::di] % ftype::expr::cj);
-    }
-    if constexpr (L + 1 < LVL) {
-      dim2lvl<CRD, L + 1>(dims, lvls, asSize);
+    if constexpr (L < LVL) {
```
cliffburdick, line 258: These changes look like the same ones from the other PR?

aartbik: Yeah (I am merging them as the other PRs go in, but they were all required to keep the example running properly)

include/matx/core/make_sparse_tensor.h
aartbik: yeah, I was not sure what to do there, but we can refine later as we go

PR #841, opened 2025-01-24
Moved sparse operator() into tensor_impl_t

cliffburdick:


include/matx/core/tensor_impl.h
```diff
@@ -57,6 +58,7 @@ struct DenseTensorData {
 
 template <typename T, typename CRD, typename POS, typename TF>
 struct SparseTensorData {
+  using value_type = T;
```
aartbik, line 61: shall we move this to L62?

(so we have sparse_data first, then the template params T/CRD/POS/TF)

include/matx/core/tensor_impl.h
```diff
@@ -894,6 +896,86 @@ MATX_IGNORE_WARNING_POP_GCC
...
+            }
+          }
+        }
+      }
+      return -1; // not found
+    }    
+
+    template <typename... Is>
+    __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ T GetSparseValue(Is... indices) const noexcept { 
+      // Element getter (viz. "lhs = Acoo(0,0);"). Note that due to the compact
```
aartbik, line 951: move this at the top of the GetSparseValue() def, rather than inside as implementation comment?

aartbik (approved):

include/matx/core/tensor_impl.h
cliffburdick: sure

PR #843, opened 2025-01-28
First version of MATX SpMM (using dispatch to cuSPARSE)

aartbik:


include/matx/operators/matmul.h
```diff
@@ -113,7 +114,17 @@ namespace matx
 
         template <typename Out, typename Executor>
         void Exec(Out &&out, Executor &&ex) const {
-          if constexpr (!std::is_same_v<PermDims, no_permute_t>) {
+          // Perform SpMM or otherwise GEMM.
+          static_assert(!is_sparse_tensor_v<OpB>, "sparse rhs not implemented");
```
cliffburdick, line 118: what's the reason this can't be done?

aartbik: We have a direct SpMM, not MSpM in cuSPARSE

include/matx/transforms/matmul/matmul_cusparse.h
```diff
@@ -0,0 +1,327 @@
...
+    static_assert(RANKB == 2);
+    static_assert(RANKC == 2);
+
+    MATX_ASSERT(a.Size(RANKA - 1) == b.Size(RANKB - 2), matxInvalidSize);
+    MATX_ASSERT(c.Size(RANKC - 1) == b.Size(RANKB - 1), matxInvalidSize);
+    MATX_ASSERT(c.Size(RANKC - 2) == a.Size(RANKA - 2), matxInvalidSize);
+
+    params_ = GetGemmParams(c, a, b, stream, alpha, beta);
+
+    cusparseStatus_t ret = cusparseCreate(&handle_);
```
cliffburdick, line 127: All the places `ret` is only used in an assert needs `[[maybe_unused]]` before it since it's disabled in release mode right now

aartbik:
> Can you add tests for these?

Yeah, I am planning a big drop with all tests, but I am still pondering a bit on the right structure

PR #844, opened 2025-01-29
Add print function with nonzero dim args

tbensonatl:
Add a new matx::print() implementation that accepts one or more arguments to specify the number of elements to print in the given dimension. This uses the pre-existing matx::fprint(stdout, ...) function.

### DIFF
--- a/test/00_operators/OperatorTests.cu
+++ b/test/00_operators/OperatorTests.cu
@@ -4749,7 +4749,7 @@ TYPED_TEST(OperatorTestsFloatAllExecs, Print)
   auto r1 = ones<TestType>(t1.Shape());
   print(r1);
 
-  auto t3 = matx::make_tensor<double>({3, 2, 20});
+  auto t3 = matx::make_tensor<TestType>({3, 2, 20});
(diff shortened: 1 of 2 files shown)


test/00_operators/OperatorTests.cu
```diff
@@ -4745,9 +4745,12 @@ TYPED_TEST(OperatorTestsFloatAllExecs, Print)
   MATX_ENTER_HANDLER();
   using TestType = cuda::std::tuple_element_t<0, TypeParam>;
 
-  auto t = make_tensor<TestType>({3});
-  auto r = ones<TestType>(t.Shape());
+  auto t1 = make_tensor<TestType>({3});
+  auto r1 = ones<TestType>(t1.Shape());
+  print(r1);
+
+  auto t3 = matx::make_tensor<double>({3, 2, 20});
```
cliffburdick, line 4752: Maybe make the type `TestType` just to be uniform with the other code?

tbensonatl: Good catch. This was originally a copy-paste from the docs. I updated to `TestType`

cliffburdick (approved):

PR #849, opened 2025-01-30
First version of MATX Sparse-Direct-Solve (using dispatch to cuDSS)

aartbik:


examples/sparse_tensor.cu
```diff
@@ -133,10 +133,38 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
...
+  // | 1 2 0 0 |   | 1 5 |   |  5 17 |
+  // | 0 3 0 0 | x | 2 6 | = |  6 18 |
+  // | 0 0 4 0 |   | 3 7 |   | 12 28 |
+  // | 0 0 0 5 |   | 4 8 |   | 20 40 |
+  //
+  // Note that X and Y are presented by reshaping a 1-dim
+  // representation of the column-major storage, because the
+  // underlying library currently only supports column-major.
+  //
+  tensor_t<float, 1> coeffs{{nse}};
```
cliffburdick, line 154: `make_tensor` should be used instead of a `tensor_t` directly in all these cases

aartbik: changed all examples

examples/sparse_tensor.cu
```diff
@@ -133,10 +133,38 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
...
+  tensor_t<int, 1> colidx{{nse}};
+  coeffs.SetVals({ 1, 2, 3, 4, 5 });
+  rowptr.SetVals({ 0, 2, 3, 4, 5 });
+  colidx.SetVals({ 0, 1, 1, 2, 3 });
+  auto Acsr = experimental::make_tensor_csr(coeffs, rowptr, colidx, {4, 4});
+  print(Acsr);
+  tensor_t<float, 1> Y{{8}};
+  tensor_t<float, 1> X{{8}};
+  // TODO: how to avoid the row-major/column-major issue?
+  Y.SetVals({5, 6, 12, 20, 17, 18, 28, 40}); // col-major
```
cliffburdick, line 144: All of our existing APIs take row-major tensors in, even if the library expects col-major. The logic to do the transposition should be inside of the transform file. Here's how we do it in SVD:
https://github.com/NVIDIA/MatX/blob/6325970ef8c0ab3abf9c873bc65a1ea42c78e215/include/matx/transforms/svd/svd_cuda.h#L493

include/matx/operators/solve.h
```diff
@@ -0,0 +1,162 @@
...
+namespace detail {
+
+template <typename OpA, typename OpB>
+class SolveOp : public BaseOp<SolveOp<OpA, OpB>> {
+private:
+  typename detail::base_type_t<OpA> a_;
+  typename detail::base_type_t<OpB> b_;
+
+  cuda::std::array<index_t, 2> out_dims_;
+  mutable detail::tensor_impl_t<typename OpA::value_type, 2> tmp_out_;
```
cliffburdick, line 51: Does the API support batching or only 2D?

include/matx/transforms/solve/solve_cudss.h
```diff
@@ -0,0 +1,286 @@
...
+    cudssMatrixType_t mtp = CUDSS_MTYPE_GENERAL;
+    cudssMatrixViewType_t mvw = CUDSS_MVIEW_FULL;
+    cudssIndexBase_t bas = CUDSS_BASE_ZERO;
+    if constexpr (TensorTypeA::Format::isCSR()) {
+      ret = cudssMatrixCreateCsr(&matA_, params_.m, params_.k, params_.nse,
+                                 /*rowStart=*/params_.ptrA2,
+                                 /*rowEnd=*/nullptr, params_.ptrA4,
+                                 params_.ptrA0, itp, dta, mtp, mvw, bas);
+    } else {
+      MATX_THROW(matxNotSupported, "cuDSS currently only supports CSR");
```
cliffburdick, line 115: Should we eventually convert automatically?

cliffburdick (changes requested):
Mostly just the change for row major

include/matx/operators/solve.h
aartbik: cuDSS supports batched operations through the operands and we should indeed extend this transformation eventually to support those!

cudssMatrixCreateBatchCsr
cudssMatrixCreateBatchDn

include/matx/transforms/solve/solve_cudss.h
aartbik: It would be nice if we could somehow cache the handles individually. But between libs, the handles are not the same (e.g. cuSparse and cuDSS CSR handles are slightly different)

examples/sparse_tensor.cu
aartbik: Let me see if I can make that work....

include/matx/transforms/solve/solve_cudss.h
cliffburdick: Sorry, I meant if the user calls `solve()` should we restrict it to CSR or should we do an async transform into a different type?

aartbik: Next, I plan to add a convert (dense-sparse/sparse-sparse) also. Perhaps we should (for now) just rely on the user to do such conversions explicitly [they can be expensive]

PR #852, opened 2025-01-30
Added documentation on integrating with existing software

cliffburdick:


docs_input/external.rst
```diff
@@ -0,0 +1,145 @@
+.. _devexternal:
+
+Interfacing With External Code and Libraries
+############################################
+
+Existing host and CUDA can can interoperate seamlessly with MatX both by using MatX primitives in existing code,
```
tylera-nvidia, line 6: `can can` typo, I think you meant "code can"

docs_input/external.rst
```diff
@@ -0,0 +1,145 @@
...
+of the operator changes. For example, changing the `o1` statment to `t1 - t2` would change the type of the operator, but using 
+templates allows the same code to exist in `foo` without changing the type. 
+
+For more information about the *operator interface*. see :ref:`_concepts`.
+
+Inside of both `foo` and `foo_kernel` all functions in the *operator interface* are available. `op(10, 1)` will return the value 
+at the 11th row and 2nd column of the operator (0-based). Using `operator()` inside of the operator will handle all the indexing 
+logic to handle the shape and stride of the operator.
+
+The last part to mention in the code is the declaration of `o1_base`. Some operator types in MatX, such as a `tensor_t`, cannot
```
tylera-nvidia, line 116: as I'm reading this I am wondering if we should make a method for tensors / operators to return this type, rather than having the users instantiate it directly. We try to abstract the types away from users in most cases, so I feel like it would be more inline with the other API usage if we instead had something like

`foo_kerenel<<<1,1>>>(o1.GetDevicHandle() )` 

I don't think this needs to be done with this PR,  but wanted to mention it as I read through this.

tylera-nvidia (approved):

docs_input/external.rst
cliffburdick: That's something we should explore, but I'm not sure if that would create extra overhead by copying it.

docs_input/external.rst
```diff
@@ -0,0 +1,145 @@
...
+  // Existing function
+  void foo(float *data);
+
+  // Create a MatX tensor in managed memory
+  auto t1 = matx::make_tensor<float>({10, 10});
+
+  // MatX processing code
+
+  // Existing code
+  foo(matx_tensor.Data());
```
tbensonatl, line 140: I assume `matx_tensor` here should be `t1`?

docs_input/external.rst
```diff
@@ -0,0 +1,145 @@
...
+  example, if a CUDA kernel is called asynchronously with a tensor as a parameter, then the tensor goes out of scope while the kernel 
+  runs, the results are undefined.
+* The *kind* of the pointer must be known to the external code. For example, if the tensor was created in device memory, the external 
+  code must access it only where device memory is accessible.
+
+If the external code supports the *dlpack* standard, the tensor's `ToDLPack()` method can be used instead to get a `DLManagedTensor` object.
+This method is much safer since all shape and ownership can be transferred.
+
+
+Passing By Object
```
tbensonatl, line 71: Should we note in this section that pass-by-value just copies the tensor metadata and not the underlying tensor allocation?

docs_input/external.rst
```diff
@@ -0,0 +1,145 @@
...
+from a MatX operator and pass it to external code by using the `Data()` method of a tensor. Note that unlike the "pass-by-object" method, 
+this method only works for tensors since general operators do not have a data pointer.
+
+Care must be taken when passing either operators or pointers to existing code to avoid bugs:
+
+* The data is only valid for the lifetime of the tensor. If the tensor goes out of scope, the data backing the tensor is invalid. For 
+  example, if a CUDA kernel is called asynchronously with a tensor as a parameter, then the tensor goes out of scope while the kernel 
+  runs, the results are undefined.
+* The *kind* of the pointer must be known to the external code. For example, if the tensor was created in device memory, the external 
+  code must access it only where device memory is accessible.
```
tbensonatl, line 65: You already mention it elsewhere, but should we reiterate here that the layout must be known by the callee in the case of passing by pointer?

docs_input/external.rst
cliffburdick: Good catch

docs_input/external.rst
cliffburdick: Added

docs_input/external.rst
tbensonatl: Good point. I think the intro of treating linear memory as a 10x10 tensor basically covers this caveat already.

tbensonatl (approved):

PR #854, opened 2025-01-31
Use C++17 RNG rather than cuRAND for host RNG

cliffburdick:


cliffburdick:
Closing this since the objects are too large on the host

PR #857, opened 2025-02-04
Move empty storage construction to inline helper method

aartbik:


include/matx/core/make_sparse_tensor.h
```diff
@@ -37,6 +37,12 @@
 namespace matx {
 namespace experimental {
 
+// Helper method to create empty storage.
+template <typename T> __MATX_INLINE__ static auto makeEmptyStorage() {
```
cliffburdick, line 41: This name is a bit deceiving because storage is a templated type as well as the type of buffer inside of the storage. This is really something like `makeDefaultNonOwningStorage`

aartbik: renamed

cliffburdick (approved):

PR #861, opened 2025-02-05
Cleanup to use pass-by-reference more consistently

aartbik:


include/matx/operators/solve.h
```diff
@@ -56,7 +56,6 @@ class SolveOp : public BaseOp<SolveOp<OpA, OpB>> {
   using matxop = bool;
   using matx_transform_op = bool;
   using solve_xform_op = bool;
-  using value_type = typename OpA::value_type;
```
cliffburdick, line 59: Every operator should have this defined. Did it need to be removed?

aartbik: Oh, this was auto formatted, I will put the instances back

include/matx/transforms/convert/sparse2dense_cusparse.h
```diff
@@ -215,33 +215,20 @@ using sparse2dense_cache_t =
 
 } // end namespace detail
 
-template <typename Op>
```
cliffburdick, line 218: What was the reason for removing this? Does cusparse support every type of input? This would include all types, strides, etc

aartbik: checking on type is still a TODO (waiting for my tests);
it was auto removed, but I put it back for later use

cliffburdick (approved):

PR #868, opened 2025-02-05
First version of MATX dense2sparse conversion (dispatch to cuSPARSE)

aartbik:


examples/sparse_tensor.cu
```diff
@@ -149,5 +149,17 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
...
+  //
+  // A direct dense2sparse conversion.
+  //
+  auto D = make_tensor<float, 2>({4, 8});
+  D.SetVals({
+    {0, 11,  0, 12,  0,  0,  0,  0},
+    {0,  0, 13,  0,  0,  0,  0,  0},
+    {0,  0,  0,  0,  0,  0,  0, 14},
+    {0, 15,  0,  0, 16,  0, 17,  0}});
+  (Acoo = dense2sparse(D)).run(exec);
```
cliffburdick, line 161: I would put a comment here with the caveats we discussed. Namely it's only possible with a direct assignment, can be slow, etc

include/matx/core/sparse_tensor.h
```diff
@@ -104,13 +150,22 @@ class sparse_tensor_t
     this->SetSparseData(v, c, p);
   }
 
-  // Default destructor.
-  __MATX_INLINE__ ~sparse_tensor_t() = default;
+  // A direct sparse tensor assignment (viz. (Acoo = ...).exec();).
+  template <typename T>
+  [[nodiscard]] __MATX_INLINE__ __MATX_HOST__ auto operator=(const T &op) {
+    return detail::sparse_set(*this, op);
```
cliffburdick, line 156: This should probably have a static assert that the RHS is a dense tensor only

cliffburdick (approved):

PR #870, opened 2025-02-06
Add "zero" matrix factor methods for COO,CSR,CSC

aartbik:


include/matx/core/make_sparse_tensor.h
```diff
@@ -37,9 +37,23 @@
 namespace matx {
 namespace experimental {
 
+// Helper method to create zero storage.
+template <typename T>
+__MATX_INLINE__ static auto
+makeDefaultNonOwningZeroStorage(index_t sz, matxMemorySpace_t space) {
+  T *ptr;
+  matxAlloc((void **)&ptr, sz * sizeof(T), space, 0);
+  for (index_t i = 0; i < sz; i++) {
```
cliffburdick, line 46: Can you add a note to replace this with something more efficient later?

cliffburdick (approved):

include/matx/core/make_sparse_tensor.h
aartbik: added

PR #871, opened 2025-02-06
Add sparse factory method tests

aartbik:


test/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ set (test_sources
     01_radar/MVDRBeamformer.cu
     01_radar/ambgfun.cu
     01_radar/dct.cu
+    sparse/Sparse.cu
```
cliffburdick, line 40: 00_sparse?

aartbik: done

test/sparse/Sparse.cu
```diff
@@ -0,0 +1,180 @@
...
+
+template <typename T> class SparseTest : public ::testing::Test { };
+
+template <typename TensorType> class SparseTestsAll : public SparseTest<TensorType> { };
+
+// For all operations that run on host and device.
+TYPED_TEST_SUITE(SparseTestsAll, MatXAllTypesAllExecs);
+
+// For all operations that run on device only.
+TYPED_TEST_SUITE(SparseTestsAllCUDA, MatXAllTypesCUDAExec);
```
cliffburdick: "AllTypes" doesn't really make sense here I don't think because it includes types like bool which aren't used for sparse. Maybe do just the ones supported by the libraries?

cliffburdick (commented):
I think "Sparse.cu" is too generic of a name if this is going in its own folder. If it's just going to be one large file it can go in 00_tensor with that name. If you plan on having a lot of separate tests, This might be "Basic.cu" or something.

test/sparse/Sparse.cu
aartbik: For the libraries, I agree. But for the factory methods it is already nice that we can handle all types (with an eye on later enhancements). For example, "bool" would make sense for an adjacency matrix representation of a graph (although for a binary-valued sparse tensor, one would not even store the values explicitly of course). And "int" would make sense for a weighted-graph.

aartbik:
> I think "Sparse.cu" is too generic of a name if this is going in its own folder. If it's just going to be one large file it can go in 00_tensor with that name. If you plan on having a lot of separate tests, This might be "Basic.cu" or something.

Renamed.

cliffburdick (approved):

PR #873, opened 2025-02-07
Add sparse conversion tests (dense2sparse, sparse2dense)

aartbik:


test/00_sparse/Basic.cu
```diff
@@ -60,10 +56,12 @@ TYPED_TEST(SparseTestsAll, MakeZeroCOO) {
   ASSERT_EQ(A.posSize(1), 0);
   ASSERT_EQ(A.crdSize(0), 0);
   ASSERT_EQ(A.crdSize(1), 0);
+  // Element getter.
+  ASSERT_EQ(A(0, 0), static_cast<TestType>(0)); // not found
```
cliffburdick, line 60: Why does this say not found?

aartbik: If you prefer, I could say something like "implicit value"?

cliffburdick: That makes sense

cliffburdick (approved):

PR #874, opened 2025-02-07
Add sparse direct-solver tests

aartbik:


include/matx/transforms/solve/solve_cudss.h
```diff
@@ -253,6 +250,8 @@ void sparse_solve_impl_trans(TensorTypeC &c, const TensorTypeA &a,
        a.Size(RANKA - 1) == b.Size(RANKB - 1) &&
        a.Size(RANKA - 2) == b.Size(RANKB - 1) &&
        b.Size(RANKB - 2) == c.Size(RANKC - 2), matxInvalidSize);
+  static_assert(std::is_same_v<typename TensorTypeA::pos_type, int32_t> &&
+		std::is_same_v<typename TensorTypeA::crd_type, int32_t>, "unsupported index type");
```
cliffburdick, line 254: Should this be `index_t` instead or is this only used internally?

cliffburdick (approved):

include/matx/transforms/solve/solve_cudss.h
aartbik: cuDSS only accepts 32-bit integers as index currently (so index_t will not work)

PR #883, opened 2025-02-21
allow for incoming and outgoing transformations on solve

aartbik:
(1) not very useful right now, since we have the surrounding
    transpose code to deal with column-major restriction;
    but this code will serve its purspose once that is removed

(2) also includes some auto formatting changes


include/matx/transforms/solve/solve_cudss.h
```diff
@@ -219,54 +220,81 @@ using gemm_cudss_cache_t =
     std::unordered_map<SolveCUDSSParams_t, std::any, SolveCUDSSParamsKeyHash,
                        SolveCUDSSParamsKeyEq>;
 
+template <typename Op>
+__MATX_INLINE__ auto getSolveSupportedTensor(const Op &in,
+                                             cudaStream_t stream) {
+  const auto func = [&]() {
+    if constexpr (is_tensor_view_v<Op>)
+      return in.Stride(Op::Rank() - 1) == 1;
+    return true;
```
cliffburdick, line 229: I think this might generate a warning with some compilers if you don't wrap in an `else`

cliffburdick (approved):

include/matx/transforms/solve/solve_cudss.h
aartbik: even with the fall into return?

cliffburdick: Yes, because it's not mutually exclusive, as written, without the `else`. It's really a compiler problem, but you could get a dead code line on the bottom return if the constexpr `if` is `true`

aartbik: Ah, I see. That makes sense (although then still here, the Stride is never fully known at compile-time, right? Still, I can change all my instances.

cliffburdick: the stride doesn't matter here. it's that the if constexpr needs that else following it to make the two paths mutually exclusive

PR #888, opened 2025-02-24
modify constexpr fall-through return into else

aartbik:


include/matx/transforms/convert/dense2sparse_cusparse.h
```diff
@@ -252,7 +252,8 @@ __MATX_INLINE__ auto getD2SSupportedTensor(const Op &in, cudaStream_t stream) {
   const auto func = [&]() {
     if constexpr (is_tensor_view_v<Op>)
       return in.Stride(Op::Rank() - 1) == 1;
-    return true;
+    else
```
cliffburdick, line 255: Please add braces around these to match the style of the other code. We had a clang formatter at one point and we need to integrate that again. Sorry for not mentioning this earlier

aartbik: Done

cliffburdick (approved):

Issue #887, opened 2025-02-21
[BUG] Poor pwelch performance


deanljohnson:
**Describe the Bug**

`pwelch` reduction stage performs about 10x worse than a similar hand-rolled CUDA implementation. 

The relevant parameters in this case are:
- fft size: 65536
- window size: 500
- window overlap: 250
- number of input samples: 164750
- number of FFTs: 658

From looking at the `pwelch_impl` source, it looks like it is effectively 3 operations:

1. FFTs
2. `X_with_overlaps = conj(X_with_overlaps) * X_with_overlaps`
3. `Pxx = sum(mag_sq_X_with_overlaps, {0}) * norm_factor`

On my system using matx, step 2 takes about 40us and step 3 takes 380us.
A custom CUDA kernel for steps 2&3 is taking 30us.

**To Reproduce**
1. Allocate `complex<float>` input tensor of size 164750
2. Create a hamming window of size 500
3. Call `matx::pwelch(input, window, 500, 250, 65536)`

**Expected Behavior**
Better performance

**Code Snippets**
The custom CUDA kernel in this case is:

```
template<typename ComplexT, typename RealT>
__global__ void ComputePowerSpectrumMultiFFT(const ComplexT* fft_buff, RealT* power_spectrum,
   int num_ffts, int num_bins)
{
   const int id = blockIdx.x * blockDim.x + threadIdx.x;
   const int num_threads = blockDim.x * gridDim.x;

   for (int bin = id; bin < num_bins; bin += num_threads)
   {
      RealT pwr = 0;
      for (int fft = 0; fft < num_ffts; fft++)
      {
         pwr += norm(fft_buff[fft * num_bins + bin]);
      }
      pwr /= num_ffts;

      // our kernel is outputting in dB, but that difference shouldn't matter.
      // Although, it would be nice if MatX allowed the `10 * log10` to be fused
      // with the rest of pwelch somehow.
      if (pwr <= 0)
         power_spectrum[bin] = 0;
      else
         power_spectrum[bin] = 10 * log10(pwr);
   }
}
```

This kernel is not doing the FFT portion of `pwelch`, but I am not reporting any issue with the FFT portion of `pwelch`, only the reduction stages.

**System Details (please complete the following information):**
 - OS: Rocky 9
 - CUDA version:CUDA 12.3
 - g++ version: 11.4.1

**Additional Context**
magnitude squared calculation info from nsys profile:

```
Begins: 14.0713s
[...]

PR #897, opened 2025-03-03
Add configurable Pwelch scaling and improve performance

tmartin-gh:
This PR addresses concerns in issue #887:

* Configurable scaling is supported.  Density and Spectrum modes, similar to Matlab or scipy.signal, as well as fused dB modes frequently needed for spectrum-analyzer-like plotting

* Custom reduction kernel is used instead of MatX sum() using CUB DeviceSegmentedReduce (DSR).

Three reasons why the custom reduction kernel is preferred:

1. In the nominal case where the input signal is a memory backed complex-valued tensor and the output power spectrum is a memory backed real-valued tensor, the intermediate overlapping signal is a 2D tensor {batches, nfft}.  This is an optimal memory layout for the FFTs, but a very suboptimal layout for CUB DSR which would the permuted layout {nfft, batches} as CUB DSR uses one block of threads to read all the 'batches' of a single fft bin.  To improve CUB performance, we'd either need to have the batched FFTs directly write to a permuted tensor (about 4x slower for the scenario considered) or permute the intermediate tensor after the FFT (extra round trip through global memory). 

2. CUB currently needs begin/end indexes for each data segment, the generation of which takes around 50% longer than the custom reduction (ignoring the permutation issue)

3. The custom reduction allows for configurable scaling, such as fusing a dB conversion.

### DIFF
--- a/examples/pwelch.cu
+++ b/examples/pwelch.cu
@@ -69,7 +69,7 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
   (x = tmp_x).run(exec); // pre-compute x, tmp_x is otherwise lazily evaluated
 
   // Create window
-  auto w = make_tensor<complex>({nperseg});
+  auto w = make_tensor<float>({nperseg});
   (w = flattop<0>({nperseg})).run(exec);
 
   // Create output tensor
--- a/examples/pwelch.cu
+++ b/examples/pwelch.cu
@@ -50,18 +50,15 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
   MATX_ENTER_HANDLER();
   using complex = cuda::std::complex<float>;
 
-  float exec_time_ms;
-  const int num_iterations = 100;
-  index_t signal_size = 256;
-  index_t nperseg = 32;
-  index_t nfft = nperseg;
-  index_t noverlap = 8;
-  float ftone = 3.0;
+  const int num_iterations = 500;
+  index_t signal_size = 256000;
+  index_t nperseg = 512;
+  index_t noverlap = 256;
+  index_t nfft = 65536;
+
+  float ftone = 2048.0;
   cudaStream_t stream;
   cudaStreamCreate(&stream);
-  cudaEvent_t start, stop;
-  cudaEventCreate(&start);
-  cudaEventCreate(&stop);
   cudaExecutor exec{stream};
 
   // Create input signal as a complex exponential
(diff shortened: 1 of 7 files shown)


examples/pwelch.cu
```diff
@@ -71,31 +68,30 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
   auto x = make_tensor<complex>({signal_size});
   (x = tmp_x).run(exec); // pre-compute x, tmp_x is otherwise lazily evaluated
 
+  // Create window
+  auto w = make_tensor<complex>({nperseg});
```
cliffburdick, line 72: Would this be a tensor usually or just passed as an operator?

include/matx/operators/pwelch.h
```diff
@@ -154,22 +170,41 @@ namespace matx
...
   template <typename xType, typename wType>
-    __MATX_INLINE__ auto pwelch(const xType& x, const wType& w, index_t nperseg, index_t noverlap, index_t nfft)
+    __MATX_INLINE__ auto pwelch(
+        const xType& x,
+        const wType& w,
+        index_t nperseg,
+        index_t noverlap,
+        index_t nfft,
+        PwelchOutputScaleMode output_scale_mode = PwelchOutputScaleMode_Spectrum,
+        typename xType::value_type::value_type fs = 1
```
cliffburdick, line 190: maybe put a static assert that it's complex so they don't get a weird error if they pass a real?

include/matx/transforms/pwelch.h
```diff
@@ -34,8 +34,71 @@
 
 namespace matx
 {
+
```
cliffburdick, line 37: This should go in include/matx/kernels

cliffburdick (approved):

examples/pwelch.cu
tmartin-gh: Can be tensor or operator

include/matx/transforms/pwelch.h
tmartin-gh: Moved

include/matx/operators/pwelch.h
tmartin-gh: Done

include/matx/kernels/pwelch.cuh
```diff
@@ -0,0 +1,101 @@
...
+          }
+          else
+          {
+            t_out(tid) = cuda::std::numeric_limits<typename T_OUT::value_type>::lowest();
+          }
+        }
+        else if constexpr (OUTPUT_SCALE_MODE == PwelchOutputScaleMode_Density_dB)
+        {
+          pxx /= (batches * fs);
+          if (pxx != 0)
```
cliffburdick, line 87: braces on an if/else usually go on the same line to match all the rest of the code. We had a clang format at one point...

cliffburdick (approved):

include/matx/kernels/pwelch.cuh
tmartin-gh: Fixed

PR #903, opened 2025-03-06
feat: added economic QR

mfzmullen:
Added an economic qr calling the cusolver orgqr methods based on appropriate type of the input. Did not replace `qr_solver` so as to keep existing functionality, but expanded on it. Modifying the existing `qr` transform to perform an economic QR seems to lead to numeric instability, and `QH * Q` is not near identity in that case, and `Q*R` is not near the input operator.  The` static_cast`s to `int` in the `orgqr_dispatch` functions prevent narrowing errors at compilation.

To catch an edge case error, I modified the hashing function slightly to use the rank of the input as part of the hash to prevent `bad any_cast` when using a cached plan. Previously an error was thrown if one firsts performs a batched economic QR on a tensor of shape` {b, m, n}` with `b = 1`, then run a plan on a tensor with shape `{m, n}` with no specified batch size. Not sure when anyone would do that except me when testing, but it is a simple error to prevent.

### DIFF
--- a/include/matx/operators/qr.h
+++ b/include/matx/operators/qr.h
@@ -184,4 +184,77 @@ __MATX_INLINE__ auto qr_solver(const OpA &a) {
   return detail::SolverQROp(a);
 }
 
+
+namespace detail {
+  template<typename OpA>
+  class EconQROp : public BaseOp<EconQROp<OpA>>
+  {
+    private:
+      typename detail::base_type_t<OpA> a_;
+
+    public:
+      using matxop = bool;
+      using value_type = typename OpA::value_type;
+      using matx_transform_op = bool;
+      using qr_solver_xform_op = bool;
+
+      __MATX_INLINE__ std::string str() const { return "qr_econ()"; }
+      __MATX_INLINE__ EconQROp(const OpA &a) : a_(a) { }    
+
+      // This should never be called
+      template <typename... Is>
+      __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) operator()(Is... indices) const = delete;
+
+      template <typename Out, typename Executor>
+      void Exec(Out &&out, Executor &&ex) {
+        static_assert(cuda::std::tuple_size_v<remove_cvref_t<Out>> == 3, "Must use mtie with 2 outputs on qr_econ(). ie: (mtie(Q, R) = qr_econ(A))");     
+
+        qr_econ_impl(cuda::std::get<0>(out), cuda::std::get<1>(out), a_, ex);
+      }
+
+      static __MATX_INLINE__ constexpr __MATX_HOST__ __MATX_DEVICE__ int32_t Rank()
+      {
+        return OpA::Rank();
+      }
+
+      template <typename ShapeType, typename Executor>
+      __MATX_INLINE__ void PreRun([[maybe_unused]] ShapeType &&shape, Executor &&ex) noexcept
+      {
+        if constexpr (is_matx_op<OpA>()) {
+          a_.PreRun(std::forward<ShapeType>(shape), std::forward<Executor>(ex));
+        }
+      }
+
+      // Size is not relevant in qr_solver() since there are multiple return values and it
+      // is not allowed to be called in larger expressions
+      constexpr __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ index_t Size(int dim) const
+      {
+        return a_.Size(dim);
+      }
+
+  };
+}
+
+/**
+ * Perform a QR decomposition on a matrix using cuSolver or a LAPACK host library.
+ * 
+ * If rank > 2, operations are batched.
+ * 
+ * @tparam OpA
+ *   Data type of input a tensor or operator
+ *
+ * @param a
+ *   Input tensor or operator of shape `... x m x n`
+ * 
+ * @return
+ *   Operator that produces R/householder vectors and tau tensor outputs.
+ *   - **Out** - Of shape `... x m x n`. The householder vectors are returned in the
+ *               bottom half and *R* is returned in the top half.
+ *   - **Tau** - The scalar factors *tau* of shape `... x min(m, n)`.
+ */
+template<typename OpA>
+__MATX_INLINE__ auto qr_econ(const OpA &a) {
+  return detail::EconQROp(a);
+}
+
 }
--- a/include/matx/operators/qr.h
+++ b/include/matx/operators/qr.h
@@ -236,7 +236,7 @@ namespace detail {
 }
 
 /**
- * Perform a QR decomposition on a matrix using cuSolver or a LAPACK host library.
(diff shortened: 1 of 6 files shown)


include/matx/operators/qr.h
```diff
@@ -184,4 +184,77 @@ __MATX_INLINE__ auto qr_solver(const OpA &a) {
...
+      constexpr __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ index_t Size(int dim) const
+      {
+        return a_.Size(dim);
+      }
+
+  };
+}
+
+/**
+ * Perform a QR decomposition on a matrix using cuSolver or a LAPACK host library.
```
cliffburdick, line 239: cuSolver in this case since you mentioned LAPACK is not supported.

mfzmullen: Thanks for the catch, updated the rest of the description too.

cliffburdick:
Thanks @mfzmullen ! This is really great work. Do you know if the instability was caused by something in particular (maybe the size or type)?

mfzmullen:
I was doing m, n = 512x256 at the time. To be honest, I did not dig into it that much. My guess (truly guess, have not checked) is in the IFELSE of `qr`, where the check is only against zero elements and not elements within a few eps of 0 could be causing some issues. Despite that, using the cusolver method was an order of magnitude faster on my setup than my modified version of the existing `qr`.

cliffburdick:
It looks like the tests are failing here:
```
[----------] 7 tests from SVDSolverTestNonHalfTypes/0, where TypeParam = cuda::std::__4::tuple<float, matx::cudaExecutor>
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDBasic
[       OK ] SVDSolverTestNonHalfTypes/0.SVDBasic (132 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDMLeqN
[       OK ] SVDSolverTestNonHalfTypes/0.SVDMLeqN (9 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDReducedMode
[       OK ] SVDSolverTestNonHalfTypes/0.SVDReducedMode (75 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDHostAlgoQR
[       OK ] SVDSolverTestNonHalfTypes/0.SVDHostAlgoQR (8 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDBasicBatched
[       OK ] SVDSolverTestNonHalfTypes/0.SVDBasicBatched (137 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/0.SVDBasicBatchedSmallMGTN
terminate called without an active exception
Aborted (core dumped)
```

I don't think you touched SVD directly and I don't see any usage of QR inside SVD. Any ideas what's causing this?

mfzmullen:
@cliffburdick updated docs, let me know if the use case is not clear. 

I am replicating that error but am not sure why it exists yet. I was using the gtest_filter to only run tests on this code, my apologies I didn't run on the full suite (I didn't expect it to cause issues elsewhere).

mfzmullen:
When I do a clean build of main and run `./matx_test --gtest_filter="*SVD*"`, I still get the error. Here it is in more detail 

> [----------] 7 tests from SVDSolverTestNonHalfTypes/4, where TypeParam = cuda::std::__4::tuple<double, matx::cudaExecutor>
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDBasic
[       OK ] SVDSolverTestNonHalfTypes/4.SVDBasic (26 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDMLeqN
[       OK ] SVDSolverTestNonHalfTypes/4.SVDMLeqN (6 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDReducedMode
[       OK ] SVDSolverTestNonHalfTypes/4.SVDReducedMode (12 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDHostAlgoQR
[       OK ] SVDSolverTestNonHalfTypes/4.SVDHostAlgoQR (5 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDBasicBatched
[       OK ] SVDSolverTestNonHalfTypes/4.SVDBasicBatched (51 ms)
[ RUN      ] SVDSolverTestNonHalfTypes/4.SVDBasicBatchedSmallMGTN
err: Failed to allocate memory. May be an asynchronous error from another CUDA call(700 != 0)
terminate called after throwing an instance of 'matx::detail::matxException'
  what():  matxException (matxOutOfMemory: ) - /home/michael/adl/forks/MatX/include/matx/core/allocator.h:197

mfzmullen:
Are there nightly tests or anything that would help track down when this started failing?

mfzmullen:
May not have time to look at this much more today, but using compute-sanitizer there seems to be an issue in the call to `cusolverDnSgesvdjBatched` in svd_cuda with float test type.

cliffburdick:
> Are there nightly tests or anything that would help track down when this started failing?

We have tests run on every commit, so it's strange this is showing up now. I will take a look at SVD since maybe it's been there all along and is just showing up with the right combination of parameters.

mfzmullen:
Pushed a commit that fixes SVD on my end. It appears(?) that not enough memory was being allocated in d_space for the batched solver (based on the NVIDIA sample [here](https://github.com/NVIDIA/CUDALibrarySamples/blob/4125fcded1ff466efe73e82e1ef98ef9572fbc84/cuSOLVER/gesvdjBatched/cusolver_gesvdjBatched_example.cu#L184))

cliffburdick:
Thanks for fixing an unrelated bug, and sorry that happened to show up here!. Merging

cliffburdick (approved):

cliffburdick:
@mfzmullen I'm seeing QR tests fail. Going to re-run this to see if this is the cause.

mfzmullen:
Thanks. I have been trying to replicate locally but no luck so far, I have one more thing to try to replicate it. Any error logs you can share?

cliffburdick:
Don't worry about it. I'll keep an eye on it and try to reproduce here.

cliffburdick:
> [Here ](https://github.com/NVIDIA/MatX/blob/650050d678527251041cea2d4ffa5fbddee999a9/include/matx/transforms/qr/qr_cuda.h#L174), why not do something akin to this? Value of eps is somewhat arbitrary, but protects against division by small numbers which may be causing some instability? As mentioned I can't reproduce the error in the first place, but this still passes on my machine.
> 
> ```
> NTypeS eps = static_cast<NTypeS>(1e-10);
> (IFELSE(abs(sc) > eps,
>              w = (v / sqrt(sc)),
>              w = NTypeS(0))).run(stream);
> ```

Given that it was not returning an error and instead just had zeros, the problem may not be related to tolerance. I'll try to reproduce it when I get time, but I'm guessing it's about the specific hardware it was running on.

PR #904, opened 2025-03-10
Add SpMV support for matvec transformation

aartbik:
with tests and doc

### DIFF
--- a/include/matx/transforms/matmul/matvec_cusparse.h
+++ b/include/matx/transforms/matmul/matvec_cusparse.h
@@ -46,22 +46,19 @@ namespace matx {
 namespace detail {
 
 /**
- * Parameters needed to execute a cuSPARSE GEMM.
+ * Parameters needed to execute a cuSPARSE SpMV.
  */
-struct MatMulCUSPARSEParams_t {
+struct MatVecCUSPARSEParams_t {
   MatXDataType_t dtype;
   MatXDataType_t ptype;
   MatXDataType_t ctype;
-  int rank;
   cudaStream_t stream;
   float alpha;
   float beta;
   index_t nse;
   index_t m;
   index_t n;
-  index_t k;
   cusparseOperation_t opA;
-  cusparseOperation_t opB;
   // Matrix handles in cuSPARSE are data specific (unlike e.g. cuBLAS
   // where the same plan can be shared between different data buffers).
   void *ptrA0;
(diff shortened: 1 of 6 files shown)


include/matx/transforms/matmul/matvec_cusparse.h
```diff
@@ -0,0 +1,323 @@
...
+    const cusparseSpMVAlg_t algo = CUSPARSE_SPMV_ALG_DEFAULT;
+    const cudaDataType comptp = MatXTypeToCudaType<TC>(); // TODO: see above
+    [[maybe_unused]] cusparseStatus_t ret =
+        cusparseSpMV(handle_, params_.opA, &salpha_, matA_, vecB_, &sbeta_,
+                     vecC_, comptp, algo, workspace_);
+    MATX_ASSERT(ret == CUSPARSE_STATUS_SUCCESS, matxMatMulError);
+  }
+
+private:
+  cusparseHandle_t handle_ = nullptr; // TODO: share handle globally?
```
cliffburdick, line 199: We typically never want to share a handle globally since it could cause unknown issues like stalls if there are locks behind using it.

include/matx/transforms/matmul/matvec_cusparse.h
```diff
@@ -0,0 +1,323 @@
...
+  cusparseDnVecDescr_t vecC_ = nullptr;
+  size_t workspaceSize_ = 0;
+  void *workspace_ = nullptr;
+  detail::MatVecCUSPARSEParams_t params_;
+  TC salpha_;
+  TC sbeta_;
+};
+
+/**
+ * Crude hash on GEMM to get a reasonably good delta for collisions. This
```
cliffburdick: GEMV

cliffburdick (approved):

include/matx/transforms/matmul/matvec_cusparse.h
aartbik: Too bad. It would save some overhead for each operation setup (not only SpMV, but all the others too)

include/matx/transforms/matmul/matvec_cusparse.h
aartbik: Fixed

include/matx/transforms/matmul/matvec_cusparse.h
cliffburdick: To be clear, this *may* work, but you'd have to ask the cusparse authors.

PR #914, opened 2025-03-12
Guard tests that require CUDSS

nvjonwong:
00_sparse/Solve.cu requires cuDSS which is not a required dependency.


cliffburdick (changes requested):
Rather than not compiling the file I think we should selectively enable/disable the tests like here: https://github.com/NVIDIA/MatX/blob/650050d678527251041cea2d4ffa5fbddee999a9/test/00_solver/Cholesky.cu#L47

This way people can see that they're missing some tests.

cliffburdick (changes requested):
I mentioned this in the last comment, but I think it's preferrable to selectively disable the test rather than not compiling the test at all.

PR #916, opened 2025-03-12
minor code cleanup

aartbik:


cliffburdick (approved):

cliffburdick:
Unit test is failing:

```
[----------] 3 tests from MatmulSparseTestsAll/1, where TypeParam = cuda::std::__4::tuple<matx::matxHalf<__nv_bfloat16>, matx::cudaExecutor>
[ RUN      ] MatmulSparseTestsAll/1.MatmulCOO
 ** On entry to cusparseSpMM_bufferSize() parameter number 6 (matB) had an illegal value: CUDA_R_16BF (__nv_bfloat16) is not supported on GPUs with compute capability < 8.0 (current: 7.5)
```

cliffburdick:
> That's strange, and certainly not related to this PR
> 
> But do we have a mixture of testing machines?
> 
> capability < 8.0 (current: 7.5)
> 
> Does MatX have a minimum capability requirement? If so, we need more guards I think

We should work on >= 7.0. Would be good to eventually see why 7.5 is not working for this.

aartbik:
Agreed. I will follow up to avoid this:

 CUDA_R_16BF (__nv_bfloat16) is not supported on GPUs with compute capability < 8.0 (current: 7.5)

Issue #920, opened 2025-03-13
[BUG] Memory leak when calling median on a flattened tensor


DanaKimball-VTS:
**Describe the Bug**
Running the median operator on a flattened device tensor results in large memory allocations and leaked memory.

**To Reproduce**
1. Allocate an input tensor with the following characteristics: 1,000,000 or more elements, element type float, rank 2 or greater, and using device memory.
2. Allocate an output tensor with the following characteristics: rank 0, using host memory, and element type float.
3. Pass the input tensor to the flatten() operator.
4. Pass the flattened tensor to the median operator and assign the result to the output tensor. Run the expression.
5. Repeat Step 4 many times.
Result: Device memory usage will climb until a matxOutOfMemory exception is thrown.

**Expected Behavior**
Memory usage may increase temporarily while performing a median operation, but it should not remain high or accumulate across multiple calls to median.

**Code Snippets**
I've attached an example program that reproduces the problem by using median or flatten independently without issue, but crashes when it attempts to run them together. To build the program, run the following commands in the directory with the main.cu and CMakeLists.txt files:
```bash
mkdir build
cd build
cmake ..
make
./mem_leak_example
```

**System Details:**
 - OS: WSL2 Ubuntu-24.02
 - CUDA version: CUDA 12.2
 - g++ version: 13.3.0
 - CMake version: 3.28.3

[Flatten_Median_Leak_Example.zip](https://github.com/user-attachments/files/19236751/Flatten_Median_Leak_Example.zip)

cliffburdick:
Hi @DanaKimball-VTS , we have a fix going through CI. It should be merged shortly.

DanaKimball-VTS:
Thank you! Greatly appreciated!

PR #922, opened 2025-03-13
Fixed bug in sort() where memory was not properly freed

cliffburdick:


cliffburdick:
Fixes https://github.com/NVIDIA/MatX/issues/920

PR #930, opened 2025-03-20
various stylistic changes to sparse tensor format file

aartbik:
(1) use uppercase template vars
(2) use cuda::std in case we want to run on device (3) make print level private
(3) fixed typo in Expr (was expr in two branches)
(4) more short-cuts for readability

### DIFF
--- a/include/matx/core/sparse_tensor_format.h
+++ b/include/matx/core/sparse_tensor_format.h
@@ -43,13 +43,13 @@ namespace experimental {
 // variables (e.g. di, di div 2, or di mod 2).
 //
 enum class LvlOp { Id, Div, Mod };
-template <LvlOp o, int d, int c> class LvlExpr {
+template <LvlOp O, int D, int C = 0> class LvlExpr {
 public:
-  static constexpr LvlOp op = o;
-  static constexpr int di = d;
-  static constexpr int cj = c;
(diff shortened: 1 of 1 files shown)


include/matx/core/sparse_tensor_format.h
```diff
@@ -165,23 +166,24 @@ template <int D, typename... S> class SparseTensorFormat {
...
+      using type0 = cuda::std::tuple_element_t<0, LvlSpecs>;
+      using type1 = cuda::std::tuple_element_t<1, LvlSpecs>;
       return type0::Expr::isId(1) && type0::Type::isDense() &&
              type1::Expr::isId(0) && type1::Type::isCompressed();
     }
     return false;
   }
 
-  template <typename CRD, int L = 0>
+  template <class CRD, int L = 0>
```
cliffburdick, line 177: We typically only use `typename` in here.

cliffburdick (approved):

include/matx/core/sparse_tensor_format.h
aartbik: Fair enough. I will change that part back.

cliffburdick: Those are likely mistakes. I'll merge this for now and clean up later.

PR #936, opened 2025-03-25
implement basic interp

### LINKED ISSUE #926: [FEA] Add interp1 operator
Interpolating from one tensor to another is a useful operator.  We should add this functionality to MatX.  Below is a description of three initial APIs i'm proposing:

//1: fully customizable option.  Caller supplies source values and coordinates along with destination coordinates
auto op = interp1<axis=rank-1>(x, v, xq, method=linear)
// axis = axis to interpolate along
// x:  tensor of input coordinates 
// v:  tensor of input values
// xq: tensor of output coordinates
// method:  interpolation method (linear, cubic, spline, etc)
// op: tensor operator of values interpolated at xq from v/x.

//2: partially deduced version.  Caller supplies input values and input locations are deduced.  Output locations are supplied by caller.  
auto op = interp1<axis=rank-1>(v, xq, method=linear) 
// axis = axis to interpolate along
// v:  tensor of input values
// xq: tensor of output coordinates
// method:  interpolation method (linear, cubic, spline, etc)
// op: tensor operator of values interpolated from v at coordinates xq.  v coordinates are [0, v.Size(axis))

//3: fully deduced version.  User supplies values.  Locations of input and output are deduced.
auto op = interp1<axis=rank-1>(v, oshape, method=linear)
// axis = axis to interpolate along
// v:  tensor of input values at locations [0, v.Size(axis))
// oshape: shape of output tensor
// op: tensor operator of values interpolated from v at coordinates [0, v.Size(axis)) to [0, oshape.Size(axis))

Implementation notes:  we should only need to implement the first version.  Deductions are implemented by supplying linspace in place of x and xq.

To start we should do method=nearest|next|previous|linear|cubic.  Then we can look into if spline makes sense.  

Later we may want to also implement interp2/interp3.

simonbyrne:
Fixes #926 

Implements linear interpolation, as well as piecewise constant (nearest, next, previous).

## Open questions
Some questions it would be good to get some input on:

### How should extrapolation be handled?

At the moment, it just extrapolates the boundary value as piecewise constant. 
- [Matlab](https://www.mathworks.com/help/matlab/ref/double.interp1.html#btwp6lt-1-extrapolation) fills with NaNs by default. 
- We could also support linear extrapolation (from the 2 points closest to the neighbor)
- numpy supports periodic domains.

### Implementation

Currently this is implemented as a regular operator, so each point is interpolated independently. There is probably room to optimize this (though the optimal pattern will probably depend on the sizes of `x` and `xq`). cc @tmartin-gh 

### Special case: equally spaced `x`

I haven't yet implemented the equally-spaced case yet, which would let us avoid the binary search. Ideally we would be able to detect if `x` is a `linspace` operator: is there a way to do this?

Similarly, we could provide a two argument method, where `x` is assumed to be `linspace(0, n-1, n)`

### What methods to support

Currently supports
- piecewise constant (`nearest`, `next`, `previous` in Matlab `interp1`)
- linear (`linear`)
- $C^2$ splines with not-a-knot boundary conditions (`spline`)

Can also add support for $C^1$ splines (`makima` and `pchip`) but I need to understand a bit more about these. Also could use different boundary conditions for splines (e.g. natural cubic splines <=> 2nd derivative equal to 0 at boundary)

### Dimensionality of samples

Currently the sample points and values must be 1d, the query points can be of arbitrary dimensions. 

We could support arbitrary dimensions sample points and values, along with an `Axis` parameter which specifies dimension the interpolation is performed on. The usual broadcasting semantics (i.e. implicit prepending of dimensions) would be compatible with what is currently implemented here.

### DIFF
--- a/examples/interpolate.cu
+++ b/examples/interpolate.cu
@@ -48,18 +48,15 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
 
 
   index_t in_count = 10;
-  auto x = make_tensor<float>({in_count});
-  auto v = make_tensor<float>({in_count});
   index_t out_count = 100;
-  auto xq = make_tensor<float>({out_count});
-  auto vq = make_tensor<float>({out_count});
 
-  (x = linspace<0>(x.Shape(), 0.0f, 1.0f)).run(exec);
-  (v = sin(x)).run(exec);
-  (xq = linspace<0>(xq.Shape(), 0.0f, 1.0f)).run(exec);
+  auto x = linspace(0.0f, 1.0f, in_count);
+  auto v = sin(x);
+  auto xq = linspace(0.0f, 1.0f, out_count);
 
   // Execute the interpolation
-  (vq = interp(x, v, xq)).run(exec);
+  auto vq = make_tensor<float>({out_count});
+  (vq = interp1<InterpMethodLinear>(x, v, xq)).run(exec);
   exec.sync();
 
   // Print the results
--- a/examples/interpolate.cu
+++ b/examples/interpolate.cu
@@ -1,0 +1,98 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2025, NVIDIA Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its
+//    contributors may be used to endorse or promote products derived from
+//    this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+/////////////////////////////////////////////////////////////////////////////////
+
+#include "matx.h"
+#include <cassert>
+#include <cstdio>
+
+using namespace matx;
+
+int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
+{
+  MATX_ENTER_HANDLER();
+
+
+  // Create a CUDA stream and executor
+  cudaStream_t stream;
+  cudaStreamCreate(&stream);
+  cudaExecutor exec{stream};
+
+
+  index_t in_count = 10;
+  auto x = make_tensor<float>({in_count});
+  auto v = make_tensor<float>({in_count});
+  index_t out_count = 100;
+  auto xq = make_tensor<float>({out_count});
+  auto vq = make_tensor<float>({out_count});
+
+  (x = linspace<0>(x.Shape(), 0.0f, 1.0f)).run(exec);
+  (v = sin(x)).run(exec);
+  (xq = linspace<0>(xq.Shape(), 0.0f, 1.0f)).run(exec);
+
+  // Execute the interpolation
+  (vq = interp(x, v, xq)).run(exec);
+  exec.sync();
+
+  // Print the results
+  std::cout << "Interpolation Results:" << std::endl;
+  std::cout << "x values: ";
+  for (int i = 0; i < 5; i++) {
+    std::cout << x(i) << " ";
+  }
+  std::cout << "..." << std::endl;
+
+  std::cout << "v values: ";
+  for (int i = 0; i < 5; i++) {
+    std::cout << v(i) << " ";
+  }
+  std::cout << "..." << std::endl;
+
+  std::cout << "xq values: ";
+  for (int i = 0; i < 5; i++) {
+    std::cout << xq(i) << " ";
+  }
+  std::cout << "..." << std::endl;
+
+  std::cout << "Interpolated values: ";
+  for (int i = 0; i < 5; i++) {
+    std::cout << vq(i) << " ";
+  }
+  std::cout << "..." << std::endl;
+
+  // Clean up
+  cudaStreamDestroy(stream);
+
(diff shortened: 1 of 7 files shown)


cliffburdick:
@simonbyrne we've had a request from a customer to add in a cubic spline interpolation. Would that be possible to add to this PR?

simonbyrne:
Yes, I'm investigating those at the moment. What is the priority here?

To clarify, are they interested in C^2 continuous splines (i.e. `interp1` with `"spline"`), or one of the C^1 continuous cubic spline methods (`"cubic"`, `"makima"`, `"pchip"`)? C^2 splines are a bit trickier, as they require a tridiagonal solve.

simonbyrne:
@cliffburdick I think this is ready for review: it would be great to get some feedback on the overall implementation (in particular the use of SFINAE), and the questions above.

examples/interpolate.cu
```diff
@@ -0,0 +1,98 @@
...
+  index_t out_count = 100;
+  auto xq = make_tensor<float>({out_count});
+  auto vq = make_tensor<float>({out_count});
+
+  (x = linspace<0>(x.Shape(), 0.0f, 1.0f)).run(exec);
+  (v = sin(x)).run(exec);
+  (xq = linspace<0>(xq.Shape(), 0.0f, 1.0f)).run(exec);
+
+  // Execute the interpolation
+  (vq = interp(x, v, xq)).run(exec);
```
cliffburdick, line 62: I think to make this clearer (if possible) it shouldn't use tensors and instead just create operators for the linspace and sin input. something like 
```
auto v = sin(x);
```

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+      // for cubic spline interpolation
+
+      using matxop = bool;
+    private:
+      O dl_, d_, du_, b_;
+      OpX x_;
+      OpV v_;
+
+    public:
+      InterpSplineTridiagonalFillOp(O dl, O d, O du, O b, OpX x, OpV v)
```
cliffburdick, line 63: Probably doesn't matter, but I think these can all be `const &` ?

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+          : dl_(dl), d_(d), du_(du), b_(b), x_(x), v_(v)  {}
+
+      __device__ inline void operator()(index_t idx) const
+      {
+      
+        if (idx == 0) { // left boundary condition
+          typename OpX::value_type x0 = x_(idx);
```
cliffburdick, line 70: May want to create aliases at the top for all these so it's not repeated. Something like:
```
using x_val_type = typename OpX::value_type;
```

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+    protected:
+      typename detail::base_type_t<OpX> x_;    // Sample points
+      typename detail::base_type_t<OpV> v_;    // Values at sample points
+      typename detail::base_type_t<OpXQ> xq_;  // Query points
+
+      mutable detail::tensor_impl_t<value_type, 1> m_; // Derivatives at sample points (spline only)
+      mutable value_type *ptr_m_ = nullptr;
+
+      template <typename... Is>
+      __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) searchsorted(const domain_type x_query) const
```
cliffburdick, line 155: Can just be `auto` since it's const

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+          cusparseHandle_t handle = nullptr;
+          cusparseStatus_t cusparse_status = cusparseCreate(&handle);
+          MATX_ASSERT(cusparse_status == CUSPARSE_STATUS_SUCCESS, matxCudaError);
+          cusparse_status = cusparseSetStream(handle, stream);
+          MATX_ASSERT(cusparse_status == CUSPARSE_STATUS_SUCCESS, matxCudaError);
+
+          size_t workspace_size = 0;
+          void* workspace = nullptr;
+          if constexpr (std::is_same_v<value_type, float>) {
+            cusparse_status = cusparseSgtsv2_bufferSizeExt(
```
cliffburdick, line 366: Does this need a check to make sure cusparse is installed? It's not a default currently.

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+              n,                  // n
+              1,                  // batch_count
+              ptr_dl_,            // sub-diagonal
+              ptr_d_,             // main-diagonal
+              ptr_du_,            // super-diagonal
+              ptr_m_,             // right-hand side and solution
+              n,                  // batch_stride
+              &workspace_size);   // workspace size
+          } else if constexpr (std::is_same_v<value_type, double>) {
+            cusparse_status = cusparseDgtsv2_bufferSizeExt(
```
cliffburdick, line 377: @aartbik added a lot of sparse support recently. I'm wondering if this functionality can be captured in the generic sparse work rather than have it specific to `interp1`?

include/matx/operators/interp.h
simonbyrne: Are there checks elsewhere? I see
```
#include <cusparse.h>
```
in several other header files.

include/matx/operators/interp.h
```diff
@@ -0,0 +1,474 @@
...
+ *   Interpolation method type (InterpMethodLinear, InterpMethodNearest, InterpMethodNext, InterpMethodPrev)
+ * @param x
+ *   Sample points (must be sorted in ascending order)
+ * @param v
+ *   Sample values (must be the same size as x)
+ * @param xq
+ *   Query points where to interpolate
+ * @returns Operator that interpolates values at query points
+ */
+template <typename Method = InterpMethodLinear, typename OpX, typename OpV, typename OpXQ>
```
cliffburdick, line 469: I think we should try to avoid the method as a template parameter in the user-facing code here. Since this is a fairly complex function anyways, I think an extra branch in `operator()` to check the method is going to be insignificant in the runtime. You can keep roughly what you have already as separate functions and just use overloads to catch the right one rather than sfinae

include/matx/operators/interp.h
cliffburdick: I just sent the wrong place. Let me fix that comment...

cliffburdick: Ok ignore my comment. I forgot cusparse comes with CTK. I was thinking of cuDSS.

include/matx/operators/interp.h
simonbyrne: Are you suggesting using an enum instead of the template parameters?

One downside of that approach is that we can't support options (e.g. for boundary conditions or extrapolation)

cliffburdick: Yes. In general we've moved away from template parameters as much as possible due to trying to match python and matlab as much as possible. I think this is one of those cases where there's enough work to do inside each branch that the branch itself would be insignificant. I was just suggesting that you could keep it as templatized internally, but you would have a single `interpolate` method that would do something like: 
```
if (method == method_a)
  interpolate<METHOD_A, ...>();
else if (method == method_b)

etc
```

simonbyrne: Okay, I can make that change.

Since it will typically be a compile-time constant, I imagine the branches should be optimized away.

include/matx/operators/interp.h
aartbik: To elaborate, the UST "encompasses" many sparse storage formats (CSR, CSC, COO, etc) and DIAG (diagonals stored as full 1-dim arrays) in included as well. So the tri-diagonal case like

```
   1  0  0 -2
   0  2  0  0 
  -1  0  3  0 
   0 -1  0  4 
```

would be stored in the UST as follows

```
---- Sparse Tensor<VAL=float,POS=int,CRD=int,DIM=2,LVL=2>
format   : ( d0, d1 ) -> ( d1 - d0 : compressed, d1 : range )
distrib  : ( d0, d1 ) -> ( )
memory   : host memory
dim      : ( 4 x 4 )
lvl      : ( 7 x 4 )
nse      : 12
pos[0]   : ( 0 3 ) #2
crd[0]   : ( -2 0 3 ) #3
values   : ( -1 -1 0 0 1.1 2.2 3.3 4.4 0 0 0 -2 ) #12
data     : 68 bytes
sparsity : 57.1429%
----
```
Note that you can easily find the diagonal [-1,-1] at diag offset -2, the diagonal [1,2,3,4] at diag offset 0, and the diagonal [-2] at diag offset 3 in the values array, all with very simple offset computations. The padding is unnecessary (i.e. the extra zeros in the non-main diagonals), but compatible with the DIA format used by many frameworks.

cliffburdick:
@simonbyrne ready to merge?

cliffburdick (approved):

simonbyrne:
Yes, I'm happy with it.

PR #938, opened 2025-03-28
Fix build issues with 32-bit indices

tbensonatl:
Fix several build issues when building with 32-bit indices (i.e., with -DMATX_BUILD_32_BIT=ON). In some cases, using 32-bit indices increases performance due to simpler index calculations. When using 32-bit index builds, the user must ensure that the total size of all tensors is <= 2^31-1.

For the 1D fft functions, we use uint64_t as the type for the fft_size parameter so that overload resolution can distinguish between calls that specify an axis (e.g., {2}) versus those that specify a length (e.g., 2). This parameter is then converted to an index_t prior to instantiating the FFTOp so that internally the fft_size is an index_t. Thus, when building with 32-bit indices, the user must ensure that any length passed to the fft function does not exceed 2^31-1.


bench/00_operators/reduction.cu
```diff
@@ -29,7 +29,7 @@ template <typename ValueType>
 void reduce_0d_matx(nvbench::state &state, nvbench::type_list<ValueType>)
 {
   // Get current parameters:
-  const long long int x_len = state.get_int64("Tensor Size");
+  const index_t x_len = static_cast<index_t>(state.get_int64("Tensor Size"));
```
cliffburdick, line 32: Does this need to call a different functions like `get_int32` for 32b indexing?

include/matx/operators/fft.h
```diff
@@ -291,9 +293,10 @@ namespace matx
    *   Normalization to apply to IFFT
    */
   template<typename OpA>
-  __MATX_INLINE__ auto ifft(const OpA &a, const int32_t (&axis)[1], uint64_t fft_size = 0, FFTNorm norm = FFTNorm::BACKWARD) {
-    auto perm = detail::getPermuteDims<remove_cvref_t<OpA>::Rank()>(axis);  
-    return detail::FFTOp(a, fft_size, perm, detail::ifft_t{}, norm);
+  __MATX_INLINE__ auto ifft(const OpA &a, const index_t (&axis)[1], uint64_t fft_size = 0, FFTNorm norm = FFTNorm::BACKWARD) {
```
cliffburdick, line 296: In places that use axis we usually just use int32 since we don't expect the rank of a tensor to go over 2^31-1.

cliffburdick (approved):

include/matx/operators/fft.h
tbensonatl: I will switch back to that convention. That should also fix the docs issue.

bench/00_operators/reduction.cu
tbensonatl: Ideally yes, but nvbench currently only has `get_int64`, `get_float64`, and `get_string`. For values greater than 2^31-1, we can either use a #if guard to enable those benchmarks only with 64-bit indexing, or we could dynamic run-time checking of the returned sizes for 32-bit indices in a constexpr branch.

PR #956, opened 2025-05-02
add argsort operator

### LINKED ISSUE #700: [FEA] Add argsort function
**Is your feature request related to a problem? Please describe.**
Wanting to sort a tensor by it's indices

**Describe the solution you'd like**
A function to sort a tensor along rows (or a given dimension) interms of indices. Should work identical to numpys `argsort`

Would look like:

```cpp
(sorted_indices = matx::argsort(tensor_in, SORT_DIR_ASC)).run();
```

**Describe alternatives you've considered**

None others

**Additional context**

None

simonbyrne:
Fixes #700

### DIFF
--- a/include/matx/transforms/cub.h
+++ b/include/matx/transforms/cub.h
@@ -2132,16 +2132,16 @@ void argsort_impl(OutputTensor &idx_out, const InputOperator &a,
     }
   }
   if (!use_a) {
-    matxAlloc((void**)&a_ptr, TotalSize(a) * sizeof(a_type), MATX_MANAGED_MEMORY, exec.getStream());
+    matxAlloc((void**)&a_ptr, TotalSize(a) * sizeof(a_type), MATX_ASYNC_DEVICE_MEMORY, exec.getStream());
     make_tensor(tmp_a, a_ptr, a.Shape());
     (tmp_a = a).run(exec);
   }
 
   // also requires a temporary for output values and input indices
-  matxAlloc((void**)&a_out_ptr, TotalSize(a) * sizeof(a_type), MATX_MANAGED_MEMORY, exec.getStream());
+  matxAlloc((void**)&a_out_ptr, TotalSize(a) * sizeof(a_type), MATX_ASYNC_DEVICE_MEMORY, exec.getStream());
   make_tensor(tmp_a_out, a_out_ptr, a.Shape());
 
-  matxAlloc((void**)&idx_in_ptr, TotalSize(idx_out) * sizeof(index_t), MATX_MANAGED_MEMORY, exec.getStream());
+  matxAlloc((void**)&idx_in_ptr, TotalSize(idx_out) * sizeof(index_t), MATX_ASYNC_DEVICE_MEMORY, exec.getStream());
   make_tensor(tmp_idx_in, idx_in_ptr, idx_out.Shape());
   (tmp_idx_in = range<RANK-1>(idx_out.Shape(), 0, 1)).run(exec);
 
--- a/include/matx/transforms/cub.h
+++ b/include/matx/transforms/cub.h
@@ -1611,7 +1611,8 @@ void sort_pairs_impl_inner(OutputIndexTensor &idx_out, const InputIndexTensor &i
     int64_t num_items = _num_items;
 #else
     if (_num_items > std::numeric_limits<int>::max()) {
-      MATX_THROW(matxInvalidSize, "Sorting is not supported for tensors with more than 2^31 items");
+      std::string err_msg = "Sorting is not supported for tensors with more than 2^" + std::to_string(std::numeric_limits<int>::digits) + " items";
+      MATX_THROW(matxInvalidSize, err_msg);
     }
     int num_segments = static_cast<int>(_num_segments);
     int num_items = static_cast<int>(_num_items);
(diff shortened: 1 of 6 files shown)


include/matx/transforms/cub.h
```diff
@@ -1540,6 +1540,133 @@ void sort_impl_inner(OutputTensor &a_out, const InputOperator &a,
...
+                                     a_in.Size(0), 0, sizeof(T1) * 8, stream);
+
+      matxFree(d_temp, stream);
+    }
+  }
+  else {
+    // CUB added support for large items and segments in https://github.com/NVIDIA/cccl/pull/3308
+    int64_t _num_segments = 1;
+    for (int i = 0; i < RANK - 1; i++) {
+      _num_segments *= a_in.Size(i);
```
cliffburdick, line 1606: You might be able to use `TotalSize(a_in);` for this

include/matx/transforms/cub.h
```diff
@@ -1540,6 +1540,133 @@ void sort_impl_inner(OutputTensor &a_out, const InputOperator &a,
...
+    }
+    int64_t _num_items = _num_segments * a_in.Size(RANK - 1);
+#if 0  // TODO: add conditional on CUB_MAJOR_VERSION once released
+    int64_t num_segments = _num_segments;
+    int64_t num_items = _num_items;
+#else
+    if (_num_items > std::numeric_limits<int>::max()) {
+      MATX_THROW(matxInvalidSize, "Sorting is not supported for tensors with more than 2^31 items");
```
cliffburdick, line 1614: Isn't it possible for `int` to be 64b on some systems and this will not trigger?

include/matx/transforms/cub.h
```diff
@@ -1949,6 +2076,127 @@ void sort_impl(OutputTensor &a_out, const InputOperator &a,
...
+  // tensor to copy the input if necessary.
+  bool use_a = false;
+  if constexpr (is_tensor_view_v<InputOperator>) {
+    if (a.IsContiguous()) {
+      make_tensor(tmp_a, a.Data(), a.Shape());
+      use_a = true;
+    }
+  }
+  if (!use_a) {
+    matxAlloc((void**)&a_ptr, TotalSize(a) * sizeof(a_type), MATX_MANAGED_MEMORY, exec.getStream());
```
cliffburdick, line 2135: We should try to avoid allocating managed memory internally. This doesn't work on WSL2 and other platforms. Can this be changed to an async device allocation?

include/matx/transforms/cub.h
```diff
@@ -1949,6 +2076,127 @@ void sort_impl(OutputTensor &a_out, const InputOperator &a,
...
+    make_tensor(tmp_a, a_ptr, a.Shape());
+    (tmp_a = a).run(exec);
+  }
+
+  // also requires a temporary for output values and input indices
+  matxAlloc((void**)&a_out_ptr, TotalSize(a) * sizeof(a_type), MATX_MANAGED_MEMORY, exec.getStream());
```
cliffburdick, line 2141: Same comments about managed

include/matx/transforms/cub.h
```diff
@@ -1949,6 +2076,127 @@ void sort_impl(OutputTensor &a_out, const InputOperator &a,
...
+  detail::tensor_impl_t<a_type, InputOperator::Rank()> tmp_a;
+  detail::tensor_impl_t<a_type, InputOperator::Rank()> tmp_a_out;
+  detail::tensor_impl_t<index_t, InputOperator::Rank()> tmp_idx_in;
+    
+  // sorting currently requires a contiguous tensor view, so allocate a temporary
+  // tensor to copy the input if necessary.
+  bool use_a = false;
+  if constexpr (is_tensor_view_v<InputOperator>) {
+    if (a.IsContiguous()) {
+      make_tensor(tmp_a, a.Data(), a.Shape());
```
cliffburdick, line 2130: More of a comment for the future, but I think a merge sort supports input iterators and not just tensors. This could eliminate the use of the temporary tensors.

cliffburdick (changes requested):
Looks like this is missing a documentation page in docs_input. I left a couple minor comments.

Looks great in general, and big thank you for doing this!

include/matx/transforms/cub.h
simonbyrne: I'm pretty sure `int` is 32-bit on all systems we support, but even if it is not, the check should still be correct (if `int` is 64-bit, then everything would just work correctly). The only thing that would be wrong in this case would be the error message.

include/matx/transforms/cub.h
simonbyrne: I did consider that, but we need the product of all dimension sizes _except_ the last one.

include/matx/transforms/cub.h
simonbyrne: It does, but I think it still mutates the keys iterator (i.e. the input tensor).

I wasn't sure exactly which sorting method to use for these. Should we make it an option?

cliffburdick: It's probably worth asking the CUB people. I know they are faster at different things and I don't recall which.

simonbyrne: It seems like the general advice is:
- for 1D, use `DeviceRadixSort`
- for ND, use `DeviceSegmentedSort`

cliffburdick: What about MergeSort?

simonbyrne: The general advice was to only use MergeSort if you need a custom operator.

In our case, we would still need allocate a temporary as merge sort mutates the input values.

cliffburdick (approved):

Issue #940, opened 2025-04-02
[FEA] Take fast path on segmented reductions


cliffburdick:
Currently if a user writes something like `(a = sum(b, {1}))` to do a sum across rows on a 2D tensor this would implicitly create an operator right now that will create an input iterator from `b`. This is not necessary on a strict unity-stride row summation, and instead we can dispatch directly to segmented functions. This PR is to address that speed issue.

cliffburdick:
Fixed in #1148

PR #959, opened 2025-05-08
Fix partial any/all reduction

### LINKED ISSUE #931: [BUG] Function "all" and "any" can not work normally
**Describe the Bug**

When the dimension of the input tensor is 3 or more, the functions all and any can't work normally.

**To Reproduce**

**Expected Behavior**

**Code Snippets**

**System Details (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - CUDA version: [e.g CUDA 11.4]
 - g++ version: [e.g. 9.3.0]

**Additional Context**

simonbyrne:
Fixes #931.


include/matx/transforms/reduce.h
```diff
@@ -804,27 +812,32 @@ template <typename T> class reduceOpMax {
...
   __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ T Init() { return (T)(0); }
+  __MATX_DEVICE__ __MATX_INLINE__ void atomicReduce(T *addr, T val) { atomicAny(addr, val); }
 };
 
 /**
- * Operator for performing an all reduction
+ * Operator for performing an any reduction
  *
- * Performs a reduction of two values of type T by returning 1 if all
+ * Performs a reduction of two values of type T by returning 1 if either
```
cliffburdick, line 827: The new comment doesn't seem right. For an `All` reduction I would except "all" or "both" values are non-zero?

cliffburdick:
Looks good! Did you happen to reproduce the original bug?

include/matx/transforms/reduce.h
simonbyrne: Ah, copy paste error.

simonbyrne:
Yes, it occurs when you do a partial reduction of a tensor rank greater than 2 (basically when CUB isn't used).

cliffburdick:
> Yes, it occurs when you do a partial reduction of a tensor rank greater than 2 (basically when CUB isn't used).

Since this code hasn't been touched in a while, is there a reason we can't use CUB in that case? We used to not use CUB in that situation before we had operators as input into CUB, now using `ReduceInput` we should support all ranks.

simonbyrne:
ah, I thought CUB required contiguous segments. Let me see if I can figure it out.

simonbyrne:
Okay, it took me a bit to figure out how it all worked, but my reading of it is that we can now use CUB for all reductions.

include/matx/transforms/reduce.h
```diff
@@ -1442,10 +1442,8 @@ void __MATX_INLINE__ reduce(OutType dest, const InType &in, ReduceOp op,
                    cudaStream_t stream = 0, [[maybe_unused]] bool init = true)
 {
   MATX_NVTX_START("", matx::MATX_NVTX_LOG_API)
-
-  constexpr bool use_cub = OutType::Rank() == 0 || (OutType::Rank() == 1 && InType::Rank() == 2);
   // Use CUB implementation if we have a tensor on the RHS and it's not blocked from using CUB
-  if constexpr (!is_matx_no_cub_reduction_v<ReduceOp> && use_cub) {
+  if constexpr (!is_matx_no_cub_reduction_v<ReduceOp>) {
```
simonbyrne, line 1446: It doesn't look like `is_matx_no_cub_reduction_v` is used any more either: is it worth deleting the old code?

cliffburdick: Yes, please delete that. I believe out reductions should not be used anymore with maybe this one as an exception:
https://github.com/NVIDIA/MatX/issues/940

simonbyrne:
Okay, I've deleted all the non-CUB reduction code.

cliffburdick (approved):

cliffburdick:
Looks like docs are failing:
```

/home/jenkins/workspace/unit-tests/docs_input/api/manipulation/selecting/reduce.rst:8: WARNING: doxygenfunction: Unable to resolve function "reduce" with arguments (OutType, TensorIndexType, const InType&, ReduceOp, cudaStream_t, bool) in doxygen xml output for project "MatX" from directory: /home/jenkins/workspace/unit-tests/build/docs_input/doxygen/xml.
Potential matches:
- template<typename InType, int D, typename ReduceOp> __MATX_INLINE__ auto reduce(const InType &in, const int (&dims)[D], ReduceOp op, bool init = true)
- template<typename InType, typename ReduceOp> __MATX_INLINE__ auto reduce(const InType &in, ReduceOp op, bool init = true)
- template<typename OutType, typename InType, typename ReduceOp> void __MATX_INLINE__ reduce(OutType dest, const InType &in, ReduceOp op, cudaStream_t stream = 0, bool init = true) [docutils]
- ```

cliffburdick:
> Ah, I see what happened. So it turns out there are two different `reduce` methods
> 
> 1. the MatX operators [in `operators/reduce.h`](https://github.com/NVIDIA/MatX/blob/4475c22efbeae70d9ccf395a686393994ff11414/include/matx/operators/reduce.h#L132-L196)
>    
>    * they don't appear in the current documentation
> 2. the mutating ones [in `transforms/reduce.h`](https://github.com/NVIDIA/MatX/blob/4475c22efbeae70d9ccf395a686393994ff11414/include/matx/transforms/reduce.h#L1317-L1454) where the output tensor is provided as an argument to the function
>    
>    * are not actually MatX operators: e.g. they accept a stream as an argument, instead of returning an operator and calling `run()`.
>    * the first of these [is mentioned in the docs](https://nvidia.github.io/MatX/api/manipulation/selecting/reduce.html#reduce)
>    * are removed by this PR
> 
> I can either remove 2 completely, or simply make it fall back on calling the operator?

Let's remove it from the docs and only have the operator

PR #961, opened 2025-05-08
Reduce FFT Memory Usage

cliffburdick:
Currently when an FFT is executed in a stream both the FFT plan and the workspace require memory. The plan uses memory for the twiddle factors, and the workspace is temporary memory used while the FFT is executing. If an application launches many FFTs this memory usage can grow significantly. 

However, if the memory is always used in the same stream, it is never accessed at the same time and is wasted. For this PR we are changing the cache manager to track workspace memory per stream, and FFTs will look up the pointer each time they run an FFT. If an FFT requiring more memory is launched in the same stream, the cache will free and re-allocate a larger block of memory. The twiddle factors can also use a noticeable amount of memory, and again they are wasted by creating many plans for the same size. We instead only allocate a single plan for all streams for a given FFT size and reuse that plan across streams. This require an extra lock on the host for multi-threaded environments, but the brief lock while making the API call should be outweighed by the memory savings.

### DIFF
--- a/include/matx/core/cache.h
+++ b/include/matx/core/cache.h
@@ -62,6 +62,11 @@ CacheId GetCacheIdFromType()
   return id;
 }
 
+struct StreamAllocation {
+  void* ptr;
+  size_t size;
+};
+
 /**
  * Generic caching object for caching parameters. This class is used for
  * creating handles/plans on-the-fly and caching them to remove the need for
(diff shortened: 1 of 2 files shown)


include/matx/core/cache.h
```diff
@@ -117,8 +122,34 @@ class matxCache_t {
...
+
+    auto el = stream_alloc_cache.find(stream);
+    if (el == stream_alloc_cache.end()) {
+      StreamAllocation alloc;
+      matxAlloc(&ptr, size, MATX_ASYNC_DEVICE_MEMORY, stream);
+      alloc.size = size;
+      alloc.ptr = ptr;
+      stream_alloc_cache[stream] = alloc;
+    }
+    else if (el->second.size < size) {
```
tbensonatl, line 136: Should we set a minimum allocation size to reduce the average number of allocations (i.e., cases such as first allocating sizeof(uint32_t), then sizeof(uint64_t), etc.)? E.g., 256 bytes or 2 MiB to correspond to the likely page size?

include/matx/transforms/fft/fft_cuda.h
```diff
@@ -286,8 +287,13 @@ template <typename OutTensorType, typename InTensorType> class matxCUDAFFTPlan_t
 
   inline void InternalExec(const void *idata, void *odata, int dir)
   {
+    // We reuse plans across streams if the plan is identical. This mutex prevents the plan from colliding across streams.
+    std::lock_guard<std::mutex> lock(mutex_);
     [[maybe_unused]] cufftResult res;
+    auto workspace = GetCache().GetStreamAlloc(params_.stream, this->workspaceSize);
```
tbensonatl, line 293: This seems to only allocate the workspace (if required) at execution time. I would expect the cache entry to still be generated at plan creation time.

include/matx/transforms/fft/fft_cuda.h
```diff
@@ -414,27 +416,22 @@ matxCUDAFFTPlan1D_t(OutTensorType &o, const InTensorType &i, cudaStream_t stream
...
                       this->params_.istride, this->params_.idist,
                       this->params_.input_type, this->params_.onembed,
                       this->params_.ostride, this->params_.odist,
                       this->params_.output_type, this->params_.batch,
-                      &workspaceSize, this->params_.exec_type);
+                      &this->workspaceSize, this->params_.exec_type);
 
   MATX_CUFFT_ASSERT_STR_EXP(error, CUFFT_SUCCESS);
 
-  matxAlloc((void **)&this->workspace_, workspaceSize, MATX_ASYNC_DEVICE_MEMORY, stream);
```
tbensonatl, line 429: See above note about still creating the cache entry at this point. We will also presumably want to call `cufftSetAutoAllocation(this->plan_, 0);` to avoid creating the workspace in the plan creation call below.

include/matx/core/cache.h
cliffburdick: Good idea. Will make it 2^21

tbensonatl (approved):

PR #962, opened 2025-05-08
simplify OptimizedExecSort

simonbyrne:
This simplifies `OptimizedExecSort` to work across multiple dimensions without batching, using the same approach as in #956.

### DIFF
--- a/include/matx/transforms/cub.h
+++ b/include/matx/transforms/cub.h
@@ -437,107 +437,40 @@ public:
 #ifdef __CUDACC__
   if constexpr (is_tensor_view_v<InputOperator>)
   {
-
-    //////////////////////////////////////////////////////
-    //////           Rank 1 Tensors               ////////
-    //////////////////////////////////////////////////////
-    if (RANK == 1)
-    {
-      if (dir == SORT_DIR_ASC)
-      {
-        cub::DeviceSegmentedSort::SortKeys(
-            d_temp, temp_storage_bytes, a.Data(), a_out.Data(),
-            static_cast<int>(a.Size(RANK-1)),
-            1,
-            BeginOffset{a}, EndOffset{a},
-            stream);
-      }
-      else
-      {
-
-        cub::DeviceSegmentedSort::SortKeysDescending(
-            d_temp, temp_storage_bytes, a.Data(), a_out.Data(),
-            static_cast<int>(a.Size(RANK-1)),
-            1,
-            BeginOffset{a}, EndOffset{a},
-            stream);
-      }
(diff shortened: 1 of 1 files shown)


include/matx/transforms/cub.h
```diff
@@ -437,107 +437,40 @@ class matxCubPlan_t {
...
+    for (int i = 0; i < RANK - 1; i++) {
+      _num_segments *= a.Size(i);
     }
-
-    //////////////////////////////////////////////////////
-    //////           Rank 2 Tensors               ////////
-    //////////////////////////////////////////////////////
-    else if (RANK == 2 || d_temp == nullptr)
+    int64_t _num_items = _num_segments * a.Size(RANK - 1);
+#if 0  // TODO: add conditional on CUB_MAJOR_VERSION once released
```
cliffburdick, line 446: What is the conditional you need here? Which CUB version supports this feature?

include/matx/transforms/cub.h
```diff
@@ -437,107 +437,40 @@ class matxCubPlan_t {
...
-        cub::DeviceSegmentedSort::SortKeysDescending(
-            d_temp, temp_storage_bytes, a.Data(), a_out.Data(),
-            static_cast<int>(a.Size(RANK-1)),
-            1,
-            BeginOffset{a}, EndOffset{a},
-            stream);
-      }
+     // CUB added support for large items and segments in https://github.com/NVIDIA/cccl/pull/3308
+    int64_t _num_segments = 1;
+    for (int i = 0; i < RANK - 1; i++) {
```
cliffburdick, line 442: Should this be `RANK - 2`?

include/matx/transforms/cub.h
simonbyrne: I don't think it's supported in a release version yet (I don't see it mentioned in https://github.com/NVIDIA/cccl/releases), but once it has we can update it.

include/matx/transforms/cub.h
simonbyrne: I don't think so? We want `i = 0, ..., RANK - 2`

cliffburdick: Maybe I'm reading this wrong, but I would expect the number of segments to be everything but the last dimension's size. So if this is looping from 0..RANK-1, it's looping over all sizes of the operator. So this is really `_num_items`?

simonbyrne: the condition is `< RANK - 1`, so `i=RANK-1` won't be included in the loop.

cliffburdick (approved):

PR #967, opened 2025-05-23
Implement DIA SpMV kernel

aartbik:
Outperforms cuSPASRE COO/CSR/ELL for DIA matrices with only a few nonzero diagonals that are near the main diagonal by quite a bit!

### DIFF
--- a/include/matx/core/make_sparse_tensor.h
+++ b/include/matx/core/make_sparse_tensor.h
@@ -38,18 +38,33 @@
 namespace matx {
 namespace experimental {
 
+template <typename T>
+__MATX_INLINE__ static void setZero(T *ptr, index_t sz,
+                                    matxMemorySpace_t space) {
+  if (space == MATX_DEVICE_MEMORY || space == MATX_ASYNC_DEVICE_MEMORY) {
+    cudaMemset(ptr, 0, sz * sizeof(T));
+  } else {
+    memset(ptr, 0, sz * sizeof(T));
+  }
+}
+
+template <typename T>
+__MATX_INLINE__ static void setVal(T *ptr, T val, matxMemorySpace_t space) {
+  if (space == MATX_DEVICE_MEMORY || space == MATX_ASYNC_DEVICE_MEMORY) {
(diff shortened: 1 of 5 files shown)


include/matx/core/make_sparse_tensor.h
```diff
@@ -38,18 +38,33 @@
...
 template <typename T>
-__MATX_INLINE__ static auto
-makeDefaultNonOwningZeroStorage(index_t sz, matxMemorySpace_t space) {
-  T *ptr = nullptr;
-  assert(sz > 0);
-  matxAlloc((void **)&ptr, sz * sizeof(T), space, 0);
+__MATX_INLINE__ static void setZero(T *ptr, index_t sz,
+                                    matxMemorySpace_t space) {
   if (space == MATX_DEVICE_MEMORY || space == MATX_ASYNC_DEVICE_MEMORY) {
     cudaMemset(ptr, 0, sz * sizeof(T));
```
cliffburdick, line 45: There's a `cudaMemsetAsync` for the async case, but you would need a stream for that.

include/matx/core/make_sparse_tensor.h
```diff
@@ -38,18 +38,33 @@
...
   } else {
     memset(ptr, 0, sz * sizeof(T));
   }
+}
+
+template <typename T>
+__MATX_INLINE__ static void setVal(T *ptr, T val, matxMemorySpace_t space) {
+  if (space == MATX_DEVICE_MEMORY || space == MATX_ASYNC_DEVICE_MEMORY) {
+    cudaMemcpy(ptr, &val, sizeof(T), cudaMemcpyDeviceToHost);
```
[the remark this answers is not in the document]
aartbik: yeah, I have no stream at this point, so opted for just syncing on async memory too.
We can refine as needed when our API's converge?

include/matx/core/make_sparse_tensor.h
aartbik: Perhaps we should (eventually) require a stream in the factory methods too.....

PR #970, opened 2025-05-30
Add axis argument to interp1.

simonbyrne:
Similar to e.g. conv1d.

Was mentioned in #926


test/00_operators/interp_test.cu
```diff
@@ -173,5 +173,86 @@ TEST(InterpTests, Interp)
   }
 
 
+  auto px2 = make_tensor<TestType>({5, 2});
```
cliffburdick, line 176: In general for next time try to break these up into multiple tests. It's a bit easier to track down bugs when we have specific test cases that fail. We're even making this more extremely recently by also creating many files for tests to increase compilation parallelism.

simonbyrne: is there a good example to copy here?

cliffburdick: btw this isn't strictly necessary, but when google test has failures it lists the name of the test at the bottom only and not the line it failed on, so you have to go searching through logs to find that. It also makes it easier to break the tests up later into separate translation units if needed.

cliffburdick (approved):

PR #972, opened 2025-05-30
Generalize DIA to DIA-I and DIA-J

aartbik:
DIA-J is most common but DIA-I is used by cuSPARSE legacy

### DIFF
--- a/examples/sparse_tensor.cu
+++ b/examples/sparse_tensor.cu
@@ -197,7 +197,7 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv) {
   auto doffsets = make_tensor<int>({3});
   dvals.SetVals({-1, -1, -1, -1, -1, 0, 4, 4, 4, 4, 4, 4, 0, 1, 1, 1, 1, 1});
   doffsets.SetVals({-1, 0, 1});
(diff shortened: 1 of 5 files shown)


examples/sparse_tensor.cu
```diff
@@ -195,8 +197,8 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv) {
   auto doffsets = make_tensor<int>({3});
   dvals.SetVals({-1, -1, -1, -1, -1, 0, 4, 4, 4, 4, 4, 4, 0, 1, 1, 1, 1, 1});
   doffsets.SetVals({-1, 0, 1});
-  auto Adia = experimental::make_tensor_dia(dvals, doffsets, {6, 6});
-  print(Adia);
+  auto AdiaJ = experimental::make_tensor_dia<false>(dvals, doffsets, {6, 6});
```
cliffburdick, line 200: It may make more sense to use a tag type here instead so it's clear what they're doing. Something like:

```
struct INDEX_I{};
struct INDEX_J{};

make_tensor_dia<INDEX_I>(...);
```

cliffburdick (approved):

PR #980, opened 2025-06-02
Vectorization and ILP

### LINKED ISSUE #975: [FEA] Vector loads/stores + ILP
For MatX's element-wise kernels it currently uses a strict model of 1 thread per value. This model is limiting for newer architecture where vector loads and stores are requires to achieve peak bandwidth. In addition to vector loads and stores, having an option for ILP can allow further speedups for memory-bound code.

cliffburdick:
Fixes #975 

This PR introduces vector loads and stores and ILP in the CUDA executor. The same code can be used in the host executor as well, but may have different alignment constraints to implement properly.

Vector loads and stores allow MatX to issue loads and stores greater than a single element. This is required to achieve higher memory bandwidth on newer GPUs where one element per thread cannot saturate the memory bandwidth of the device. The goal of the feature is to be transparent to the user; all existing code should run at either the same speed or faster than before by recompiling. 

ILP is a joint feature to vector loads and stores where the executor can specify an ILP level (`ELEMENTS_PER_THREAD`) of powers of 2 up to 32 elements per thread. Currently the EPT cannot exceed the vector width of 16 bytes in the CUDA executor (float4), but larger values in the future will force multiple loads.

Along with these features. a "capabilities" interface is also introduced that allows operators to specific their capabilities for different features. The `ELEMENTS_PER_THREAD` capability is a "min" capability, meaning the minimum value of any operator in the expression forces all the other operators to use that value. If one operator does not support more than 1 element per thread while all the others do, the entire expression will revert to 1 element per thread. To reduce the size of this PR, only a subset of operators allow multiple elements per thread, and the rest can be added as needed over time. The most common operators, tensors, binary, and unary arithmetic operators, all support the maximum vector width.


deanljohnson:
Does this change require custom operators written outside of `matx` to change? It am trying to update to a newer version of matx and am getting compile errors on my previously implemented operators. It seems like my operators are being called with an `ElementsPerThread` value. If all operators need to be updated, can the documentation in the developers guide for operators be updated? https://nvidia.github.io/MatX/developer_guide/operators.html

cliffburdick:
> Does this change require custom operators written outside of `matx` to change? It am trying to update to a newer version of matx and am getting compile errors on my previously implemented operators. It seems like my operators are being called with an `ElementsPerThread` value. If all operators need to be updated, can the documentation in the developers guide for operators be updated? https://nvidia.github.io/MatX/developer_guide/operators.html

Hi Dean, yes, unfortunately it does require a small change. In your custom operator you copy this:
https://github.com/NVIDIA/MatX/blob/3d60b056b902043fd4c07a481145996c09feca72/include/matx/operators/fftshift.h#L94-L103

Then change your operator() to have new signatures like this:
https://github.com/NVIDIA/MatX/blob/3d60b056b902043fd4c07a481145996c09feca72/include/matx/operators/fftshift.h#L58-L74

I will update the docs.

PR #987, opened 2025-06-05
add basic quaternion support from CUTLASS

simonbyrne:
This adds the most minimal required support for quaternions by aliasing quaternion from CUTLASS. As such it only works if CUTLASS support is enabled.

It mostly works out of the box: I tested basic arithmetic operators and math functions (e.g. `exp`).

Some open questions:
- The [constructor takes the arguments `(imag_i, imag_j, imag_k, real)`](https://github.com/NVIDIA/cutlass/blob/b244379d9b15574e07b73b814b88bd2233f0b3ce/include/cutlass/quaternion.h#L115). This is different than [Eigen, which uses `(real, imag_i, imag_j, imag_k)`](https://eigen.tuxfamily.org/dox/classEigen_1_1Quaternion.html#a3eba7a582f77a8f30525614821d7056f). Is this going to be a problem?
- Do we want to import the small matrix/vector types: these are used in some of the [geometric rotation functions](https://github.com/NVIDIA/cutlass/blob/b244379d9b15574e07b73b814b88bd2233f0b3ce/include/cutlass/quaternion.h#L236C22-L236C28)?


cliffburdick:
Looks good! Was it too much work to break the dependency of CUTLASS?

cliffburdick:
> Can you expand on what you mean? the class is defined in CUTLASS, so I don't see how you could break the dependency. Or do you mean copy their implementation into MatX?

Right, when we talked I thought we discussed how the cutlass implementation is so minimal and somewhat boilerplate what we could just copy the small pieces we needed into our own type and not rely on the external dependency.

cliffburdick:
> Ah, I misunderstood. I thought you meant typealias the code.
> 
> The challenge with that is that it depends on the small matrix/vector types which are also defined in CUTLASS. I'll look to see how coupled they are, perhaps it is feasible to split them out.
> 
> If we do define our own quaternion type, would we still be able to use CUTLASS for quaternion GEMMs?

I think given that the customers who care about extra dependencies are not the same customers that care about quaternions at this point, what you have is good. We can reassess in the future if needed.

cliffburdick:
Is this ready for review? It still says draft.

simonbyrne:
@cliffburdick It's functional, but probably needs more tests: what do you suggest?

There are still the open questions above, and what functionality we want (e.g. GEMM, abs, etc)

PR #990, opened 2025-06-07
revert to old complex-scalar multiplication logic

simonbyrne:
This reverts the complex-scalar multiplication logic that was used before #980.


simonbyrne:
This seems to fix the build issues I was seeing.

cliffburdick (commented):
I think we don't want to revert this because it was masking a casting issue that shouldn't be there. we shouldn't allow complex<T> * R where R is anything but a type T

PR #991, opened 2025-06-09
Fixing another 12.9 compiler bug

cliffburdick:


simonbyrne:
Thanks, I saw this as well.

PR #1006, opened 2025-07-08
Fix host compiler errors when using -Wall -Werror

tmartin-gh:
These compiler errors exhibit with matx header included in .cpp file when compiled with -Wall -Werror
* error: typedef 'using NTypeS = typename matx::inner_op_type_t<typename ATensor::value_type>::type' locally defined but not used
* error: ignoring '#pragma unroll ' [-Werror=unknown-pragmas]
* error: unused variable 'ret' [-Werror=unused-variable]

This MR provides the fix by eliminating the unused type and replacing #pragma unroll with a MATX_LOOP_UNROLL macro.

### DIFF
--- a/include/matx/transforms/fft/fft_fftw.h
+++ b/include/matx/transforms/fft/fft_fftw.h
@@ -85,9 +85,9 @@ struct FftFFTWParams_t {
   {
     MATX_NVTX_START("", matx::MATX_NVTX_LOG_INTERNAL)
     FftFFTWParams_t params;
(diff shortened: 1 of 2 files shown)


tmartin-gh:
Note: Also see instances of `MatX/include/matx/core/tensor_desc.h:263: error: ignoring '#pragma unroll ' [-Werror=unknown-pragmas]` with -Wall -Werror in .cpp files but this MR doesn't address those.

PR #1011, opened 2025-07-17
fix flipud for 1D tensors

simonbyrne:
The current `flipud` logic caused a compilation error when using it with 1D arrays. Changing the logic to seems to fix the issue. Added tests.

### DIFF
--- a/include/matx/operators/reverse.h
+++ b/include/matx/operators/reverse.h
@@ -196,12 +196,8 @@ namespace matx
   template <typename T1>
   auto __MATX_INLINE__ flipud(const T1 &t)
   {
-    if constexpr (T1::Rank() == 1)
-    {
-      return detail::ReverseOp<T1::Rank() - 1 , T1>(t);
-    }
-
-    return detail::ReverseOp<T1::Rank() - 2, T1>(t);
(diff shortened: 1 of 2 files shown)


cliffburdick (approved):

include/matx/operators/reverse.h
```diff
@@ -196,12 +196,8 @@ namespace matx
...
   auto __MATX_INLINE__ flipud(const T1 &t)
   {
-    if constexpr (T1::Rank() == 1)
-    {
-      return detail::ReverseOp<T1::Rank() - 1 , T1>(t);
-    }
-
-    return detail::ReverseOp<T1::Rank() - 2, T1>(t);
+    constexpr int dim = std::max(0, T1::Rank() - 2);
+    return detail::ReverseOp<dim, T1>(t);
```
simonbyrne, line 200: It looks like
```suggestion
    if constexpr (T1::Rank() == 1)
    {
      return detail::ReverseOp<T1::Rank() - 1 , T1>(t);
    }
    else {
      return detail::ReverseOp<T1::Rank() - 2, T1>(t);
    }
```
also works.

PR #1018, opened 2025-07-25
Remove runtime checks on memory spaces

aartbik:
(1) allow a mix of memory spaces inside the sparse tensor;
        ultimately, only the visibility matters, not the exact consistency

    (2) remove the runtime check on offset = {-1,0,+1} for now;
        I see no other solution than copying this to host or having
        some kernel check on device, which all seems too costly;
        given the very limited usage of UST, we just assume the solve
        is only called on tridiag with these offsets


aartbik:
Unrelated failure?

matxException (matxOutOfMemory: ) - /home/jenkins/workspace/unit-tests/include/matx/core/allocator.h:

cliffburdick (approved):

PR #1033, opened 2025-08-12
Add zipvec operator

tbensonatl:
Add a new zipvec operator that zips multiple input operators into a vectorized operation (e.g., an operator with type float3 when zipping x, y, and z coordinates). 

The input operators must have the same rank and size in all dimensions and the types must be compatible where compatible means that we can convert each input to the derived output type without narrowing. For types char, short, int, long (and their unsigned variants) and float/double, the vector size can be 1-4 elements. For __half, only length-2 vectors are supported. The components from the input operators are accessed by the fields x, y, z, and w, respectively, in the zipped operator.

### DIFF
--- a/include/matx/core/tensor_impl.h
+++ b/include/matx/core/tensor_impl.h
@@ -1045,14 +1045,6 @@ MATX_IGNORE_WARNING_POP_GCC
         constexpr int EPT_int = static_cast<int>(EPT);
         if constexpr (EPT == detail::ElementsPerThread::ONE) {
           return data_.ldata_[GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...))];
-        } else if constexpr (sizeof(T) != detail::alignment_by_type<T>()) {
-          // If the alignment of the type does not match the types size, then we use the Vector load
-          // method. That method specializes this case to load the elements as scalars in a loop over
-          // all EPT elements. A common case where sizeof(T) != detail::alignment_by_type<T>() is for vector types,
-          // e.g. float3 has an alignment of 4B, but a size of 12B.
-          detail::Vector<T, EPT_int> vec;
-          vec.template load<EPT_int>(data_.ldata_ + GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...)));
-          return vec;
         } else if constexpr (EPT_int * sizeof(T) <= MAX_VEC_WIDTH_BYTES ) {
           return *reinterpret_cast<detail::Vector<T, EPT_int>*>(data_.ldata_ + GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...)));
         } else {
--- a/include/matx/core/tensor_impl.h
+++ b/include/matx/core/tensor_impl.h
@@ -1045,7 +1045,15 @@ MATX_IGNORE_WARNING_POP_GCC
         constexpr int EPT_int = static_cast<int>(EPT);
         if constexpr (EPT == detail::ElementsPerThread::ONE) {
           return data_.ldata_[GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...))];
-        } else if constexpr (EPT_int * detail::alignment_by_type<T>() <= MAX_VEC_WIDTH_BYTES ) {
+        } else if constexpr (sizeof(T) != detail::alignment_by_type<T>()) {
+          // If the alignment of the type does not match the types size, then we use the Vector load
+          // method. That method specializes this case to load the elements as scalars in a loop over
+          // all EPT elements. A common case where sizeof(T) != detail::alignment_by_type<T>() is for vector types,
+          // e.g. float3 has an alignment of 4B, but a size of 12B.
+          detail::Vector<T, EPT_int> vec;
+          vec.template load<EPT_int>(data_.ldata_ + GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...)));
+          return vec;
+        } else if constexpr (EPT_int * sizeof(T) <= MAX_VEC_WIDTH_BYTES ) {
           return *reinterpret_cast<detail::Vector<T, EPT_int>*>(data_.ldata_ + GetValC<EPT, 0, Is...>(cuda::std::make_tuple(indices...)));
         } else {
           detail::Vector<T, EPT_int> vec;
--- a/include/matx/core/tensor_impl.h
+++ b/include/matx/core/tensor_impl.h
@@ -1173,17 +1173,25 @@ MATX_IGNORE_WARNING_POP_GCC
           return detail::ElementsPerThread::ONE;
         }
 
-        int width = MAX_VEC_WIDTH_BYTES / alignment_by_type<T>();
-        while (width > 1) {
-          if (((Lsize() % width) == 0) &&                                       // Last dim is a multiple of vector load size
-            ((reinterpret_cast<uintptr_t>(data_.ldata_) % (alignment_by_type<T>() * width)) == 0)) {
-            break;
+        if constexpr (sizeof(T) != alignment_by_type<T>()) {
+          // If the alignment of the type does not match sizeof(T), then the logic
+          // below will not necessarily work. It would generally work if alignment_by_type<T>() < sizeof(T),
+          // except for cases where sizeof(T) > MAX_VEC_WIDTH_BYTES, which result in division by zero.
+          // Curently, this is only true for vector types (e.g., double3) and we use one EPT for these cases.
+          return detail::ElementsPerThread::ONE;
+        } else {
+          int width = MAX_VEC_WIDTH_BYTES / sizeof(T);
+          while (width > 1) {
+            if (((Lsize() % width) == 0) &&                                       // Last dim is a multiple of vector load size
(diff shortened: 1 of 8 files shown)


include/matx/core/vector.h
```diff
@@ -57,7 +87,7 @@ struct alignas(sizeof(T) * N) Vector {
   // loads need to be issued to fill the vector
   template <int EPT>
   __MATX_INLINE__ __MATX_HOST__ __MATX_DEVICE__ void load(T* ptr) {
-    constexpr int elements_per_load = (MAX_VEC_WIDTH_BYTES / sizeof(T));
+    constexpr int elements_per_load = (MAX_VEC_WIDTH_BYTES / alignment_by_type<T>());
```
cliffburdick, line 90: I'm not clear on how this works. Given a `double3` it's a 24B load (which we don't have), but the alignment is 8B. Elements per load would be 2, which to me means it will load 2 `double3` elements. However, the code right below assumes a `float4`/16B load, which doesn't work if you have a single `double3`, for example.

include/matx/core/type_utils.h
```diff
@@ -1185,6 +1185,169 @@ constexpr cusparseIndexType_t MatXTypeToCuSparseIndexType() {
...
+
+// Helper to check if all types in a pack are the same types
+template <typename... Ts>
+struct all_same;
+
+template <typename T>
+struct all_same<T> : std::true_type {};
+
+template <typename T, typename U, typename... Rest>
+struct all_same<T, U, Rest...> : std::bool_constant<std::is_same_v<T, U> && all_same<U, Rest...>::value> {};
```
cliffburdick, line 1268: I think this is just 
`cuda::std::conjunction_v<cuda::std::is_same<T, Ts>...>;`

include/matx/core/type_utils.h
```diff
@@ -1185,6 +1185,169 @@ constexpr cusparseIndexType_t MatXTypeToCuSparseIndexType() {
...
+  template <typename T, typename TargetType>
+  static constexpr bool is_non_narrowing_conversion() {
+    // Same type is always safe
+    if constexpr (std::is_same_v<T, TargetType>) {
+      return true;
+    }
+    // Integer to larger or equal integer is safe
+    else if constexpr (std::is_integral_v<T> && std::is_integral_v<TargetType>) {
+      return sizeof(T) <= sizeof(TargetType) && 
+             std::is_signed_v<T> == std::is_signed_v<TargetType>;
```
cliffburdick, line 1305: Is it still narrowing if the signs are different? It seems like having 1 sign bit vs extending the word by 8, 16, or 32 bits wouldn't affect the narrowing?

cliffburdick: I suppose that only applies if the number is positive, so ignore my comment

include/matx/core/type_utils.h
```diff
@@ -1185,6 +1185,169 @@ constexpr cusparseIndexType_t MatXTypeToCuSparseIndexType() {
...
+      if constexpr (std::is_same_v<TargetType, float>) {
+        return sizeof(T) <= 2; // 16-bit or smaller integers
+      } else if constexpr (std::is_same_v<TargetType, double>) {
+        return sizeof(T) <= 4; // 32-bit or smaller integers
+      } else {
+        return false; // Conservative for other floating point types
+      }
+    }
+    // __half conversions
+    else if constexpr (std::is_same_v<T, __half> || std::is_same_v<TargetType, __half>) {
```
cliffburdick, line 1324: Should this be comparing `__half` directly or the MatX alias for it?

include/matx/operators/zipvec.h
```diff
@@ -0,0 +1,236 @@
+////////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, NVIDIA Corporation
```
cliffburdick, line 4: 2025

include/matx/operators/zipvec.h
```diff
@@ -0,0 +1,236 @@
...
+        else return "";
+      }
+
+      __MATX_INLINE__ std::string str() const {
+        return get_str<-1>();
+      }
+
+      __MATX_INLINE__ ZipVecOp(const Ts&... ts) : ops_(ts...)
+      {
+        static_assert(sizeof...(Ts) > 0 && sizeof...(Ts) <= 4, "Must have between 1 and 4 tensors for zipvec");
```
cliffburdick, line 79: Tensors or operators?

include/matx/core/type_utils.h
```diff
@@ -1185,6 +1185,169 @@ constexpr cusparseIndexType_t MatXTypeToCuSparseIndexType() {
...
+  // Check for mixed __half with other types and provide helpful error message
+  static_assert(!has_mixed_half,
+    "zipvec does not support mixing __half with other types. "
+    "Use explicit type conversion (e.g., as_type<float>()) to convert __half operands to a common type before calling zipvec.");
+
+  // Helper to determine the common type
+  using common_type_impl = std::conditional_t<
+    all_half,
+    __half,  // If all types are __half, keep as __half
+    std::common_type_t<Ts...>  // Otherwise use standard common type (no __half mixed in due to static_assert)
```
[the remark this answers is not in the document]
tbensonatl: Updated uses of common_type, conditional_t, is_integral_v, is_signed_v, and is_floating_point_v to use the cuda:: namespace. Left std::is_same_v as that is still widely used elsewhere and likely needs a batch update.

include/matx/core/type_utils.h
tbensonatl: We could support unsigned to signed conversion when the target type is larger than the source type (just not signed to unsigned). Do we already have narrowing logic elsewhere? The issue that I ran into was that template deduction type conversion rules were more lax than braced-initialization narrowing conversion rules. But maybe I can get around this with casts in the braced initialization of the vector type too.

include/matx/core/type_utils.h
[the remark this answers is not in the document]
tbensonatl: We do have a __nv_bfloat162 type, so we could support 2D sizes. For the matxHalf<> types, I assume we would only support those as input operators? i.e., I don't think that matxHalf<__half2> or matxHalf<__nv_bfloat162> would work.

include/matx/operators/zipvec.h
tbensonatl: updated

include/matx/core/type_utils.h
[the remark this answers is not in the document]
cliffburdick: Right, this is for the inputs. We don't support `__half` as an input in general since many of the standard unary and binary ops are not defined.

include/matx/core/type_utils.h
tbensonatl: I removed this along with the is_narrowing_conversion helper. The operator now allows some narrowing conversions like int -> float, but those are the conversions supported by cuda::std::common_type.

include/matx/core/vector.h
tbensonatl: Ah, I found it in the same tensor_impl.h function from the below comment. The assumption there seems to be that any EPT * sizeof(T) value exceeding 32B will be divisible by 16B. That seems to be true for the fundamental types and our current vectorization capability enum.

include/matx/core/type_utils.h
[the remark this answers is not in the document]
tbensonatl: I removed half support from zipvec

include/matx/core/tensor_impl.h
```diff
@@ -1173,10 +1173,10 @@ MATX_IGNORE_WARNING_POP_GCC
           return detail::ElementsPerThread::ONE;
         }
 
-        int width = MAX_VEC_WIDTH_BYTES / sizeof(T);
+        int width = MAX_VEC_WIDTH_BYTES / alignment_by_type<T>();
         while (width > 1) {
           if (((Lsize() % width) == 0) &&                                       // Last dim is a multiple of vector load size
-            ((reinterpret_cast<uintptr_t>(data_.ldata_) % (sizeof(T) * width)) == 0)) {
+            ((reinterpret_cast<uintptr_t>(data_.ldata_) % (alignment_by_type<T>() * width)) == 0)) {
```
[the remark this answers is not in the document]
tbensonatl: I special-cased this logic so that the sizeof(T) != alignment_by_type<T>() returns ElementsPerThread::ONE and this logic just uses sizeof(T).

include/matx/operators/zipvec.h
```diff
@@ -0,0 +1,236 @@
...
+        // All ops must have the same scalar value type; that is enforced by AggregateToVecType
+
+        for (int32_t i = 0; i < RANK; i++) {
+            MATX_ASSERT_STR(((ts.Size(i) == pp_get<0>(ts).Size(i)) && ...), 
+                matxInvalidSize, "zipped operators must have the same size in all dimensions");
+        }
+      }
+
+      template <ElementsPerThread EPT, typename... Is>
+      __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) operator()(Is... is) const
```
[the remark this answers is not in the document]
tbensonatl: I left this for now as decltype(auto) for const seemed to be the most common pattern in the other operators. Maybe we can change this via a batch update?

cliffburdick (approved):

PR #1037, opened 2025-08-13
Fixed issue where op=transform was double-calling transform

cliffburdick:
When a statement `op = transform` occurred where op was not a tensor, the transform was being called twice incorrectly leading to poor performance. This simplifies the code and calls it once.


tbensonatl (approved):
reshape.h needs to be added, but otherwise this change looks good to me.

Issue #48, opened 2021-11-24
[FEA] Add custom allocator interface


cliffburdick:
**Is your feature request related to a problem? Please describe.**
MatX currently takes raw non-owned pointers, smart unowned pointers, and self-allocates owned pointers otherwise. While this allows for many different options, it does not allow users to have their own memory allocators MatX would call.

**Describe the solution you'd like**
Allow functionality similar to xtensor

**Describe alternatives you've considered**
Above

-

cliffburdick:
Requirements for this feature are:

1. Users can pass in a raw pointer that has been allocated, along with a custom allocator, and the allocator will call the deallocate function when the tensor goes out of scope.
2. Users can pass in a custom allocator and allocation and deallocation would be performed by MatX with that allocator
3. Both cases above must have stream and non-stream-oriented semantics

This will also allow future dynamic creation of tensors by using the underlying allocator when a copy is performed. 

Some API functions do asynchronous allocations as part of normal operation, but these do not need to call the custom allocator since it's purely managed internally by MatX and the caching layer.

At this time there are at least three popular allocators we want to support: RMM, thrust, and libcu++. libcu++ memory resource API has not been finalized at this time, so we are waiting to see how similar it would be to RMM. Either way, we will need some kind of wrapper or adapter due to the API differences.

leofang:
> 3. Both cases above must have stream and non-stream-oriented semantics
 
I assume this means users can pass in either a stream-ordered or non-ordered allocator? Then, if both are set which one should MatX pick? How should the allocator getter/setter APIs look like? (We can discuss this offline as I am also working on this for other projects 🙂)

cliffburdick:
this allocator will be used whenever a new tensor is created, and for simplicity some users won't want to specify a stream or even know what it is. the default allocator (cudaMalloc) is a good example of this. for more advanced users they would want a stream allocator bound to a tensor instance. this means that whenever a tensor is deep copied you'd expect that to happen in the specified stream, and when a tensor goes out of scope the free is stream-oriented.

we should definitely connect offline.

cliffburdick:
Since libcudacxx's implementation of the memory allocator has been deferred, it's worthwhile at this point to use RMM as our de facto memory allocator instead of the custom one we currently have. This would also give us more flexibility by using pool and arena allocators.

cliffburdick:
Closed as part of #1046

PR #1046, opened 2025-08-27
Refactor Storage system to use duck-typed allocators

### LINKED ISSUE #1043: [FEA] Remove allocator from the template type
**Is your feature request related to a problem? Please describe.**
I'm trying to add a way to use [Holoscan allocators](https://docs.nvidia.com/holoscan/sdk-user-guide/components/resources.html#allocator) with matx tensors.

I was able to do this by defining my own replacement for `raw_pointer_buffer` (see https://gist.github.com/simonbyrne/ab772dde60df14908626e614143590cd), but I hit the issue that the resulting tensor type is no longer a `matx::tensor_t<T,N>` (as it's not compatible with `DefaultStorage`), which makes it difficult to use them interchangably.

**Describe the solution you'd like**
If the allocator is removed from the template type, then we could define our own allocators by subclassing while keeping the interface consistent. 

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

cliffburdick:
- Replace template-based Storage with unified Storage<T> class
- Remove inheritance requirement from pmr_allocator_interface
- Implement pure duck typing - users just need allocate()/deallocate() methods
- Add SFINAE detection for both allocator objects and pointers
- Support owning/non-owning constructors with optional custom allocators
- Eliminate dynamic allocation overhead for common allocator cases
- Maintain backward compatibility with legacy sparse tensor types
- Add comprehensive unit tests for duck-typed allocator system
- Organize tests in new test/00_misc/ directory following MatX conventions

This provides maximum flexibility while removing inheritance requirements and virtual function overhead. Users can now implement any allocator without inheriting from base classes - if it has allocate() and deallocate() methods, it works automatically via SFINAE detection.

Closes #1043 #48

### DIFF
--- a/docs_input/api/creation/tensors/make.rst
+++ b/docs_input/api/creation/tensors/make.rst
@@ -22,14 +22,19 @@ Return by Value
 .. doxygenfunction:: make_tensor( T *data, ShapeType &&shape, bool owning = false)
 .. doxygenfunction:: make_tensor( TensorType &tensor, typename TensorType::value_type *data, typename TensorType::shape_container &&shape)
 .. doxygenfunction:: make_tensor( TensorType &tensor, typename TensorType::value_type *ptr)
-.. doxygenfunction:: make_tensor( Storage &&s, ShapeType &&shape)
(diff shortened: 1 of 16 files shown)


cliffburdick:
Passing CI but issue with the coverage uploading at the end.

PR #1155, opened 2026-04-16
add python deps to devcontainer

simonbyrne:
Python is required for building the tests, so makes sense to install them in the devcontainer dockerfile.


cliffburdick (changes requested):
The Dockerfile is not used and instead is generated by hpccm. I have a PR almost ready to merge that addresses this. I'd rather get away from a virtual environment in the container that people may not know about and instead install packages globally.

simonbyrne:
Okay, that's fair.

PR #1159, opened 2026-04-23
add emulated double->fltflt cast

simonbyrne:
This speeds up `double` to `fltflt` conversions on fp64-decoupled hardware. I also tweaked the cast benchmark so that it isn't affected by fp64 arithmetic perf.

# L40S results

## Before (with update benchmark)
```
Benchmark       float        double       fltflt       fltflt vs dbl
------------------------------------------------------------------
cast2fltflt     1.00x        32.52x       4.39x        7.40x

--------------------------------------------------------------------------------
Raw timings (auto-scaled units):

Benchmark       float           double          fltflt          fltflt vs dbl
---------------------------------------------------------------------------
cast2fltflt     618.045 us      20.100 ms       2.716 ms        7.40x
```

## After
```
Benchmark       float        double       fltflt       fltflt vs dbl
------------------------------------------------------------------
cast2fltflt     1.00x        10.65x       4.40x        2.42x

--------------------------------------------------------------------------------
Raw timings (auto-scaled units):

Benchmark       float           double          fltflt          fltflt vs dbl
---------------------------------------------------------------------------
cast2fltflt     618.190 us      6.585 ms        2.721 ms        2.42x
```

### DIFF
--- a/bench/00_misc/fltflt_arithmetic.cu
+++ b/bench/00_misc/fltflt_arithmetic.cu
@@ -1077,7 +1077,13 @@ __global__ void iterative_cast2fltflt_kernel(fltflt* __restrict__ result, int64_
         acc[ilp] = static_cast<fltflt>(src_val);
         asm volatile("" : "+f"(acc[ilp].hi), "+f"(acc[ilp].lo));
       }
-      src_val = src_val + static_cast<T>(0.0001);
+      // For double, increment the bit pattern to get the next representable value
+      // so the loop anti-aliasing doesn't introduce a double-precision add.
+      if constexpr (std::is_same_v<T, double>) {
+        src_val = __longlong_as_double(__double_as_longlong(src_val) + 1LL);
+      } else {
+        src_val = src_val + static_cast<T>(0.0001);
+      }
     }
 
     fltflt result_val = acc[0];
--- a/bench/00_misc/fltflt_arithmetic.cu
+++ b/bench/00_misc/fltflt_arithmetic.cu
@@ -1079,7 +1079,7 @@ __global__ void iterative_cast2fltflt_kernel(fltflt* __restrict__ result, int64_
       }
       // For double, increment the bit pattern to get the next representable value
       // so the loop anti-aliasing doesn't introduce a double-precision add.
-      if constexpr (std::is_same_v<T, double>) {
+      if constexpr (cuda::std::is_same_v<T, double>) {
         src_val = __longlong_as_double(__double_as_longlong(src_val) + 1LL);
       } else {
         src_val = src_val + static_cast<T>(0.0001);
--- a/include/matx/kernels/fltflt.h
+++ b/include/matx/kernels/fltflt.h
@@ -58,8 +58,56 @@ struct alignas(8) fltflt {
     // nvcc will warn about __host__ and __device__ annotations on default constructors because default
     // constructors will not run in all conditions (e.g., in static shared memory CUDA kernel allocations).
     __MATX_INLINE__ fltflt() = default;
-    __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ constexpr explicit fltflt(double x)
-        : hi(static_cast<float>(x)), lo(static_cast<float>(x - static_cast<double>(hi))) {}
+    // On device, we avoid double-precision subtraction by extracting the residual mantissa
+    // bits directly from the IEEE 754 representation using integer operations. hi is the
+    // truncated (round-toward-zero) float, so hi + lo == x exactly in real arithmetic.
+    // Ties-to-even and NaN/Inf are not handled precisely (edge cases fall back to
+    // hi=(float)x, lo=0). The constexpr branch is taken during compile-time evaluation.
+    __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ constexpr explicit fltflt(double x) {
+#if defined(__CUDA_ARCH__)
+        // During compile-time (constexpr) evaluation on device, fall through to the
+        // standard double-subtraction path which is constexpr-compatible.
+        if (__builtin_is_constant_evaluated()) {
+            hi = static_cast<float>(x);
+            lo = static_cast<float>(x - static_cast<double>(hi));
+        } else {
+            // Fast runtime path: extract residual mantissa bits via integer operations,
+            // avoiding double-precision subtraction. hi uses truncation (round-toward-zero),
+            // so hi + lo == x exactly in real arithmetic. NaN/Inf/subnormal doubles and
+            // doubles outside float range are not handled precisely (per design).
+            unsigned long long xbits = __double_as_longlong(x);
+            unsigned int sign = (unsigned int)(xbits >> 63);
+            unsigned int e_x = (unsigned int)((xbits >> 52) & 0x7FFU);
+            unsigned long long mant = xbits & 0x000FFFFFFFFFFFFFULL;
+            // float biased exponent: (e_x - 1023) + 127 = e_x - 896
+            int hi_exp = (int)e_x - 896;
+            if (e_x == 0 || hi_exp <= 0 || hi_exp >= 255) {
+                hi = (float)x;
+                lo = 0.0f;
+            } else {
+                // hi: top 23 explicit mantissa bits, same exponent as x.
+                hi = __int_as_float((sign << 31) | ((unsigned int)hi_exp << 23) | (unsigned int)(mant >> 29));
+                // lo: lower 29 bits of significand, scaled to the correct power of 2.
+                unsigned int r = (unsigned int)(mant & 0x1FFFFFFFU);
+                if (r == 0) {
+                    lo = 0.0f;
+                } else {
+                    int n = 31 - __clz(r);           // highest set bit of r (0..28)
+                    int lo_exp = n + (int)e_x - 948; // n + (e_x - 1023 - 52) + 127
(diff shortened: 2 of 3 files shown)


simonbyrne:
This will be slower on 100-class hardware, but this isn't where `fltflt` is intended to be used.

simonbyrne:
This improves the overflow/underflow handling adds a fast2sum step to keep the invariant `fl(hi + lo) == hi`. Unfortunately it adds some cost, but is still 2x faster than the old conversion.

```

Benchmark       float        double       fltflt       fltflt vs dbl
------------------------------------------------------------------
cast2fltflt     1.00x        16.85x       4.36x        3.87x

--------------------------------------------------------------------------------
Raw timings (auto-scaled units):

Benchmark       float           double          fltflt          fltflt vs dbl
---------------------------------------------------------------------------
cast2fltflt     622.532 us      10.489 ms       2.713 ms        3.87x

```

bench/00_misc/fltflt_arithmetic.cu
```diff
@@ -1077,7 +1077,13 @@ __global__ void iterative_cast2fltflt_kernel(fltflt* __restrict__ result, int64_
         acc[ilp] = static_cast<fltflt>(src_val);
         asm volatile("" : "+f"(acc[ilp].hi), "+f"(acc[ilp].lo));
       }
-      src_val = src_val + static_cast<T>(0.0001);
+      // For double, increment the bit pattern to get the next representable value
+      // so the loop anti-aliasing doesn't introduce a double-precision add.
+      if constexpr (std::is_same_v<T, double>) {
```
cliffburdick: We should try to use `cuda::std` here especially since it's a device function, even though it doesn't matter for this particular function

test/00_misc/FloatFloatTests.cu
```diff
@@ -294,6 +294,20 @@ struct FltFltCmpGe {
   }
 };
 
+struct DoubleToFltFlt {
```
cliffburdick, line 297: Should we have these defined in cast.h instead?

cliffburdick (approved):

include/matx/kernels/fltflt.h
```diff
@@ -58,8 +58,56 @@ struct alignas(8) fltflt {
...
+            hi = static_cast<float>(x);
+            lo = static_cast<float>(x - static_cast<double>(hi));
+        } else {
+            unsigned long long xbits = __double_as_longlong(x);
+            unsigned int sign = (unsigned int)(xbits >> 63);
+            unsigned int e_x = (unsigned int)((xbits >> 52) & 0x7FFU);
+            unsigned long long mant = xbits & 0x000FFFFFFFFFFFFFULL;
+            // hi_exp: float biased exponent = (e_x - 1023) + 127 = e_x - 896.
+            int hi_exp = (int)e_x - 896;
+            if (e_x == 0 || hi_exp <= 0 || hi_exp >= 255) {
```
tbensonatl, line 89: We may need a bit more guard for hi_exp (e.g., >= 254 for this path). We could end up with hi == FLT_MAX and then have that overflow to Inf during the fast2sum. That would then change our accuracy guarantees in the case that  hi_exp == 254. Or it may be enough to just add (hi_exp == 254 && hi_mantissa == 0x7FFFFF).

tbensonatl:
> This improves the overflow/underflow handling adds a fast2sum step to keep the invariant `fl(hi + lo) == hi`. Unfortunately it adds some cost, but is still 2x faster than the old conversion.

2x faster is still great! One potential concern could be increased register pressure, but it seems unlikely for that to result in the old behavior being faster. I can try it in the recently merged sarbp example to see if there is any change in register usage or any spilling for that kernel for one sanity check.

bench/00_misc/fltflt_arithmetic.cu
simonbyrne: that's fine (thought we do use this elsewhere in the repo)

include/matx/kernels/fltflt.h
simonbyrne: I actually think that would be ok: the only way it would overflow was if lo_raw >= ulp(FLT_MAX), in which case hi should round up to Inf anyway. the lo will be NaN, but that's what it would be now.

tbensonatl: I think we can have lo_raw >= ulp(FLT_MAX) (thus the need for fast2sum to renormalize). r is in [0, 2^29-1], but r_float can round up to 2^29. The scale factor in this case is the float with biased exponent 254-52=202, so the unbiased exponent is 202-127=75. Then lo_raw is 2^29 * 2^75 = 2^104 = ulp(FLT_MAX). So, in this case it's the hi + lo_raw during renormalization that will overflow s to Inf, but we generally need the renormalization because the rounding for r_float can round up to ulp(hi).

simonbyrne: actually i realized that isn't quite right. let me think about this a bit more

simonbyrne:
@tbensonatl we don't actually have the invariant `fl(hi + lo) == hi` currently: as an example, if you have the double `x` with bit pattern `0x3ff0000010000001` (1.000000059604645): this is exactly
```
x = 1 + (float eps) / 2 + (double eps)
```
`hi` will round up to the bit pattern `0x3f800001` 
```
hi = 1 + (float eps)
```
and `lo` will round down to `0xb3800000`
```
lo = - (float eps) / 2
```
and so `hi + lo = 1 + (float eps)/2`, which will round to `fl(hi + lo) = 1`.

tbensonatl:
> @tbensonatl we don't actually have the invariant `fl(hi + lo) == hi` currently:  ...
> and so `hi + lo = 1 + (float eps)/2`, which will round to `fl(hi + lo) = 1`.

Good point. This implies that we also need a fast_two_sum normalization in the standard double->fltflt conversion. We will likely want to offer both conversion approaches (the current fp64 path + normalization and your approach in this PR).

simonbyrne:
I made a few improvements, and improved the error bounds slightly.

bench/00_misc/fltflt_arithmetic.cu
simonbyrne:
```suggestion
      if constexpr (cuda::std::is_same_v<T, double>) {
```

test/00_misc/FloatFloatTests.cu
simonbyrne: These are just accessors to the internals. I guess we could consider exposing them, but that's probably beyond the scope of this PR.

simonbyrne:
@cliffburdick can you rerun build?

include/matx/kernels/fltflt.h
```diff
@@ -58,8 +58,49 @@ struct alignas(8) fltflt {
...
-    __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ constexpr explicit fltflt(double x)
-        : hi(static_cast<float>(x)), lo(static_cast<float>(x - static_cast<double>(hi))) {}
+    // On device, extract hi/lo from IEEE-754 double bits with no FP64 instructions.
+    //
+    // Accuracy guarantees (device fast path, for normal doubles in float range):
+    //   - fl(hi + lo) == hi  (fast2sum ensures no rounding error when adding lo back).
+    //   - |x - (hi+lo)| <= 8 ulp(x) if hi and lo in normal range.
+    //   - NaN, Inf, subnormal doubles, and doubles outside float range fall back to
+    //     hi = (float)x.
+    //   - If hi is NaN or Inf, lo can be arbitrary.
```
tbensonatl, line 68: Isn't lo always 0 with this implementation when hi is NaN/+Inf/-Inf? If so, we should document that as it is a tighter bound. Also, to better handle subnormals, can we just add a case for hi_exp == 0 below that uses the FP64 path? It should be rare and relatively low overhead in general since we already have a conditional.

include/matx/kernels/fltflt.h
```diff
@@ -58,8 +58,49 @@ struct alignas(8) fltflt {
...
+    __MATX_HOST__ __MATX_DEVICE__ __MATX_INLINE__ constexpr explicit fltflt(double x) {
+#if defined(__CUDA_ARCH__)
+        // Constexpr evaluation on device uses the standard double-subtraction path.
+        if (__builtin_is_constant_evaluated()) {
+            hi = static_cast<float>(x);
+            lo = static_cast<float>(x - static_cast<double>(hi));
```
tbensonatl: Please add the fast2sum normalization here as you noted that we can currently violate our fl(lo + hi) = hi invariant

include/matx/kernels/fltflt.h
tbensonatl: Sorry, the above should say e_x == 0 (the stored exponent is all zeros) for the denormals, no hi_exp == 0.

simonbyrne: > Isn't lo always 0 with this implementation when hi is NaN/+Inf/-Inf? 

i think the only case where it can be something else is for the very narrow set of values which are in the highest `float` binade, but are large enough round up to `Inf` when converting to `float`. In that case, I think you will end up with `lo` being `NaN`.

Note that the old path will also give `NaN`s.

simonbyrne: `double` subnormals (`e_x == 0`) will underflow to zero when converted to a `float`. For `float` subnormals (`hi_exp == 0`), we go via the cast-to-float path which (since `lo == 0.0`, as we can't represent anything smaller).

PR #1161, opened 2026-04-24
Add fast-path TensorAccessor & Optimize SarBp/ChannelizePoly

tbensonatl:
Adds detail::TensorAccessor (a per-kernel wrapper with an IsUnitStride fast path) and plumbs it through the SAR BP and channelize_poly transforms. Extends the SAR BP shared-mem preamble to float and mixed precisions. Adds a CTILE=32 variant of ChannelizePoly1D_SmemTiled for num_channels <= 32, which replaces the generic ChannelizePoly1D kernel for most small-channel oversampled configs (M=32/D=16, M=40/D=20 etc.) at roughly 3x the throughput. Adds a Full filter smem layout (single copy at smem[p*M + phase]) alongside the existing Rotated layout; dispatcher picks the smaller footprint. Refactors the SmemTiled FIR loops to use a single running filter_idx counter, saving 3 registers/thread. Scopes all channelize_poly internals under matx::detail::cpoly, dropping redundant MATX_CHANNELIZE_POLY1D_ and matxChannelizePoly1DInternal_ prefixes. Adds support for num_channels = 1 to channelize_poly() for completeness, although such cases should use conv1d() for better performance.

The TensorAccessor class improves performance for tensors with unit stride by eliding the load and multiply of the last-dimension stride. Speedups range from 3-12%, depending on the GPU and precision. The float path for the sarbp example improved by > 3x on GPUs with reduced double-precision throughput, but that is due to amortizing the cost of double-to-float conversions.

General polyphase channelizer improvements from the TensorAccessor class range from 0-10%. Oversampled cases with <= 48 channels improved by > 2x due to the addition of a tiled shared memory implementation with a smaller tiling factor.

### DIFF
--- a/include/matx/kernels/channelize_poly.cuh
+++ b/include/matx/kernels/channelize_poly.cuh
@@ -436,8 +436,8 @@ __global__ void ChannelizePoly1D_SmemTiled(
             }
         } else {
             // The dispatch must not select FilterInSmem when K exceeds the
-            // rotations[] array size. Assert this invariant at runtime.
-            assert(K <= detail::cpoly::SmemTiledMaxRotations);
+            // rotations[] array size. We rely on the transform dispatch to
+            // ensure this invariant due to the cost of run-time kernel checks.
             int32_t rotations[detail::cpoly::SmemTiledMaxRotations];
             for (int32_t k = 0; k < K; k++) {
                 rotations[k] = static_cast<int32_t>((static_cast<int64_t>(k) * decimation_factor) % M);
--- a/include/matx/kernels/channelize_poly.cuh
+++ b/include/matx/kernels/channelize_poly.cuh
@@ -42,22 +42,27 @@
 #include "matx/core/utils.h"
 #include "matx/core/type_utils.h"
 #include "matx/core/tensor_utils.h"
+#include "matx/kernels/tensor_accessor.h"
 #include <cuda/std/__algorithm/min.h>
 #include <cuda/std/__algorithm/max.h>
 
 namespace matx {
 
-// detail constants that require both host and device visibility at compile time
+// detail constants that require both host and device visibility at compile
+// time. Scoped to matx::detail::cpoly so the transform's helpers can sit in
+// the same namespace without colliding with other transforms' internals.
 namespace detail {
+namespace cpoly {
     // Number of output elements generated per thread
-    constexpr index_t CHANNELIZE_POLY1D_ELEMS_PER_THREAD = 1;
+    constexpr index_t ElemsPerThread = 1;
 
     // Maximum number of filter rotations per channel for the SmemTiled kernel.
-    // This is used to determine if the filter can be stored in shared memory. The
-    // number of rotations can exceed this value, but the filter will be read from
-    // global memory rather than cached in shared memory.
-    constexpr int MATX_CHANNELIZE_POLY1D_SMEM_TILED_MAX_ROTATIONS = 32;
-}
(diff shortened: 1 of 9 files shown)


include/matx/kernels/channelize_poly.cuh
```diff
@@ -388,23 +397,48 @@ __global__ void ChannelizePoly1D_SmemTiled(
...
-                    ? filter.operator()(static_cast<index_t>(h_ind))
+                    ? filter_acc(static_cast<index_t>(h_ind))
                     : static_cast<filter_t>(0);
             }
         } else {
             // The dispatch must not select FilterInSmem when K exceeds the
             // rotations[] array size. Assert this invariant at runtime.
-            assert(K <= detail::MATX_CHANNELIZE_POLY1D_SMEM_TILED_MAX_ROTATIONS);
-            int32_t rotations[detail::MATX_CHANNELIZE_POLY1D_SMEM_TILED_MAX_ROTATIONS];
+            assert(K <= detail::cpoly::SmemTiledMaxRotations);
```
cliffburdick: Do we want to use assert in device code? I thought it has a first-time and register penalty.

tbensonatl: Good point. It should be removed with NDEBUG defined, but we can rely on the transform dispatch checks to keep this from impacting debug build performance too much.

tbensonatl: Removed the assert()

include/matx/kernels/tensor_accessor.h
```diff
@@ -0,0 +1,290 @@
...
+                MATX_LOOP_UNROLL
+                for (int d = 0; d < NumOuter; ++d) {
+                    outer_strides_[d] = op.Stride(d);
+                }
+            }
+        }
+    }
+
+    // Rank-0 access.
+    template <int R = Rank, cuda::std::enable_if_t<R == 0, int> = 0>
```
cliffburdick, line 119: Since it's C++20 we should use `requires` when possible now

include/matx/transforms/channelize_poly.h
```diff
@@ -322,34 +500,54 @@ inline void matxChannelizePoly1DInternal_FusedChan(OutType o, const InType &i,
...
-    case 5:
-      ChannelizePoly1D_FusedChan<THREADS, 5, OutType, InType, FilterType, AccumType><<<grid,THREADS,0,stream>>>(o, i, filter);
-      break;
-    case 6:
-      ChannelizePoly1D_FusedChan<THREADS, 6, OutType, InType, FilterType, AccumType><<<grid,THREADS,0,stream>>>(o, i, filter);
-      break;
-    default:
-      MATX_THROW(matxInvalidDim, "channelize_poly: channel count not support with fused kernel");
+
+  constexpr bool fast_path_eligible =
```
cliffburdick, line 508: Should this be || or do they all need to be eligible?

cliffburdick (approved):

include/matx/transforms/channelize_poly.h
tbensonatl: They need to all be eligible. This is less general than handling each tensor independently, but it lets us use a single template parameter instead of three for the "normal" case that the inputs are all regular dense tensors.

include/matx/kernels/tensor_accessor.h
tbensonatl: Good point. I'll add similar support to the resample_poly kernels, so I can switch from SFINAE to requires in that PR.

PR #1172, opened 2026-05-06
Add FPAN-based version of fltflt addition with reduced latency

tbensonatl:
Switch to a floating-point accumulation network based approach for fltflt addition as presented in "High-Performance Branch-Free Algorithms for Extended-Precision Floating-Point Arithmetic" by Zhang and Aiken. This formulation uses the same number of floating point operations (20), but has a shorter critical path (10 vs 13) and thus higher ILP.

Added a new fltflt_add latency benchmark. The throughput of fltflt_add() did not change because the number of operations remains the same, but the latency reduced by ~17-18% in the most latency-exposed benchmark.

Also removed the separate fltflt and sarbp benchmark scripts and replaced them with a shorter benchmark script that leverages structured output from nvbench (JSON) rather than parsing the nvbench standard output.


bench/00_misc/fltflt_arithmetic.cu
```diff
@@ -704,6 +784,7 @@ void fltflt_bench_madd(nvbench::state &state, nvbench::type_list<PrecisionType>)
   constexpr int block_size = 256;
   int grid_size = static_cast<int>((size + block_size - 1) / block_size);
 
+  warmup_gpu_once();
```
cliffburdick, line 787: Isn't this the job of nvbench?

tbensonatl: It is. I was seeing high variability in the first benchmark run (add_throughput with floats). Adding that anecdotally stabilized the first run, but I can dig into it a bit more. nvbench will definitely handle single-run warmup (loading/copying the kernel, etc.), but didn't seem to be handling clock ramp up.

cliffburdick (approved):

PR #1177, opened 2026-05-09
Add Windows (MSVC + nvcc) build support

zcsizmadia:
# Add Windows (MSVC + nvcc) build support

Enable building MatX on Windows with MSVC and CUDA Toolkit. All examples, tests, and benchmarks compile and link. Changes are guarded behind `if(WIN32)` / `if(MSVC)` / `#ifdef _MSC_VER` and do not affect Linux (GCC/Clang) builds.

## Major obstacles

~~1. **nvbench / fmt**: nvbench is built as a shared library by default but lacks `__declspec(dllexport)` annotations, causing 59 unresolved-symbol linker errors on Windows. Additionally, fmt 11.2.0 (a transitive dependency of nvbench) contains `char32_t` literals above `U+10FFFF` that nvcc's CUDA frontend cannot parse. Workarounds: build nvbench as a static library on Windows, and pre-register fmt with a patch that caps the offending values.~~

2. **Windows SDK `ERROR` macro / `NOGDI`**: The Windows SDK header `wingdi.h` defines `#define ERROR 0`, which collides with MatX's `enum` value `matxError_t::ERROR`. Defining `NOGDI` before `<windows.h>` prevents the GDI portion of the SDK from being included, avoiding the macro conflict without needing to `#undef ERROR` throughout the codebase.

3. **MSVC warnings treated as errors**: MatX compiles with `/W4 /WX`, which surfaces dozens of warnings that GCC/Clang either don't emit or classify differently — signed/unsigned mismatches (C4245), double-to-float truncation (C4305), potentially uninitialized variables (C4701), unused parameters (C4100), unreachable code (C4702), and more. Each required a targeted, portable fix (casts, value-initialization, `[[maybe_unused]]`, etc.) rather than blanket suppression. MSVC 18 (VS 2026 Insiders) introduced additional new warnings not present in VS 2022, including C4864 (dependent-name template calls on free function templates) and C4459 (local declaration hides global, newly triggered inside Google Test headers), further expanding the set of errors that had to be addressed.

## Build environment

| Component | Version |
|-----------|---------|
| OS | Windows 11 |
| Visual Studio | 2026 (VS 18 Insiders), MSVC 14.51 |
| CUDA Toolkit | 13.2 |
| CMake | 3.31+ |
| Python | 3.14 |

Minimum supported: VS 2026 ~~VS 2019 16.11+ (same as MatX's C++20 requirement). Compatible with VS 2022~~.

## Local build steps

Open a **Visual Studio x64 Native Tools Command Prompt** (or a PowerShell/cmd session initialized with `vcvars64.bat`) to ensure MSVC, the Windows SDK, and the CUDA Toolkit are all on `PATH`.

```bash
cd MatX
mkdir build
cd build
cmake -DMATX_BUILD_TESTS=ON -DMATX_BUILD_BENCHMARKS=OFF -DMATX_BUILD_EXAMPLES=ON -DMATX_BUILD_DOCS=OFF -DCMAKE_CUDA_ARCHITECTURES=86 ..
cmake --build . --config Release
```

## Changes

- **POSIX-only API replacements**: `setenv`/`unsetenv` → `_putenv_s`/`_putenv`; `std::aligned_alloc` → `_aligned_malloc`/`_aligned_free`; `ssize_t` → `ptrdiff_t` or `int64_t`; `%lu` → `%zu` for `size_t` format strings.
- **Hardcoded POSIX paths replaced**: `/tmp/` → `std::filesystem::temp_directory_path()` for portable temporary file creation.
- **Narrowing conversion fixes** (C4267, C4244, C4305, C4245): Wrapped numeric literals and `size_t` expressions in `static_cast<TargetType>(...)` at assignment, initialization, and comparison sites.
- **Signed/unsigned mismatch fixes** (C4018, C4389): Added `static_cast` in ASSERT/loop expressions where `unsigned` test types were compared against signed integer literals.
- **Uninitialized variable fixes** (C4700, C4701): Value-initialized local arrays and structs with `{}` (`ninds{}`, `gind{}`, `shape{}`, `children_aggregated_val{}`).
- **Unused variable fixes** (C4189): Removed unused local variables or added `(void)var;` where the variable is only referenced as a template argument (e.g., `constexpr BATCH_DIMS` inside a local struct).
- **`[[nodiscard]]` discard suppression** (C4834): Wrapped intentional comma-operator `.run()` patterns with `MATX_IGNORE_WARNING_PUSH_MSVC(4834)` / `MATX_IGNORE_WARNING_POP_MSVC`.
- **Unsafe operation suppressions** (C4804, C4146): Suppressed MSVC warnings around intentional `bool`-in-division and unary-minus-on-unsigned in operator macros via `MATX_IGNORE_WARNING_PUSH/POP_MSVC`.
- **GTest header warnings scoped to tests** (C4459, C4018, C4389): Added `/wd4459` in `test/CMakeLists.txt` only; fixed test-side signed/unsigned mismatches in `ASSERT_LE` calls.
- **COFF object file section limit** (C1128): Added `/bigobj` to test and benchmark targets for heavily-templated translation units.
- **Windows SDK macro conflicts** (`NOGDI`, `NOMINMAX`, `_USE_MATH_DEFINES`): Added compile definitions to prevent `ERROR`/`max`/`min` macro collisions and provide `M_PI`/`M_E` on MSVC.
- **MSVC conformance flags**: Added `/Zc:preprocessor`, `/Zc:__cplusplus`, ~~`/utf-8` globally; `-DFMT_UNICODE=0`~~ for nvcc CUDA frontend ~~(which doesn't inherit `/utf-8` from `-Xcompiler`)~~.
- **Third-party / SDK header warning suppressions**: Globally suppressed `/wd4996` (deprecated SDK APIs), `/wd4100` (Python headers), `/wd4864` (VS18 dependent-name warning), `/wd4702` (constexpr-if unreachable code), `/wd4324` (intentional alignas padding), `/wd4714` (large templated `__forceinline`).
- **nvbench / fmt compatibility**: Built nvbench as static library on Windows (no `__declspec(dllexport)`); pre-registered fmt 11.2.0 patch for nvcc `char32_t` parse error.
- **Visibility and ABI macros**: Added `MATX_VISIBILITY_DEFAULT` macro (GCC/Clang attribute, empty on MSVC); replaced bare `__attribute__` uses.
- **`enum` to `enum : index_t`**: Prevented C4309 truncation-of-constant warnings on MSVC enum comparisons.
- **Unreachable-code removal**: Removed dead `default:` cases in switch statements that MSVC flags under `/W4`.

### Main library (`CMakeLists.txt`) — global, applies to all targets

These suppressions are applied via `CMAKE_CUDA_FLAGS` and `add_compile_options` and therefore affect the library, tests, examples, and benchmarks. All originate in external headers that cannot be modified in MatX.

| Warning | Description | Root cause | Scope |
|---------|-------------|-----------|-------|
| C4996 | Use of deprecated API | NVIDIA SDK headers (`cusparse.h`, `cusolverDn.h`) annotate types with `[[deprecated]]` | CUDA only |
| C4100 | Unreferenced formal parameter | Python 3.14 `cpython/unicodeobject.h` has intentional `_unused_op` parameters | CXX + CUDA |
| C4864 | Dependent-name template call on free function | VS18 new warning; triggers on correct code in libcudacxx / MatX operator templates | CXX + CUDA |
| C4702 | Unreachable code | `if constexpr` branches where every arm returns, but a trailing `return` is still required by MSVC for non-`constexpr` compilation paths | CXX + CUDA |
| C4324 | Structure padded due to `alignas` | Intentional padding in CUDA operator types | CXX + CUDA |
| C4714 | `__forceinline` function not inlined | Expected for large templated functions such as `get_type_str()` | CXX + CUDA |

### `test/` — additional suppressions scoped to test executables only

| Warning | Description | Root cause |
|---------|-------------|-----------|
| C4459 | Local declaration hides outer scope name | Triggered inside Google Test's `gtest-matchers.h`; not fixable in MatX |
| C4244 | Conversion from `int64_t` to `float`/`double`, possible loss of data | Intentional `(TestType)(Size() * Size() * …)` casts throughout typed tests; also triggered by GTest macro internals when comparing mixed integer/float types |

### `bench/` — no additional suppressions

Benchmark targets use only `/W4 /WX /bigobj` plus the global suppressions above. No bench-specific `wd` flags are needed.

## TODO (before merge)

- [ ] Review all warning fixes and propose better or more generic solutions where applicable (e.g. upstream fixes in MatX headers rather than suppression flags, shared helper macros, etc.)
- [ ] Approve the hidden warnings
~~- [ ] Identify a better strategy to resolve the nvbench/fmt version issues~~
- [ ] Check building on Linux gcc and clang
~~- [ ] Check building using Visual Studio 2022~~
~~- [ ] Check builidng using Visual Studio 2019 (?)~~
- [ ] Check building using CUDA Toolkit 12.x
- [ ] Confirm all tests pass
- [ ] Confirm all bench targets execute without errors
- [ ] Compact comments after changes are discussed and approved


cmake/patches/fmt-11.2.0-nvcc-char32-fix.patch
```diff
@@ -0,0 +1,35 @@
...
++++ b/include/fmt/format.h
+@@ -3161,8 +3161,25 @@ constexpr auto fractional_part_rounding_thresholds(int index) -> uint32_t {
+   // It is equal to ceil(2^31 + 2^32/10^(k + 1)).
+   // These are stored in a string literal because we cannot have static arrays
+   // in constexpr functions and non-static ones are poorly optimized.
+-  return U"\x9999999a\x828f5c29\x80418938\x80068db9\x8000a7c6\x800010c7"
+-         U"\x800001ae\x8000002b"[index];
++
++  // while in C++23 we can use static constexpr, and in c++17 we can use out of
++  // function definition of inline constexpr, in C++11 we have to rely on string
```
cliffburdick: MatX is C++20 and above only now, so we should not have to worry about what C++11 supports.

zcsizmadia: I just used the diff from https://github.com/fmtlib/fmt/commit/03cfb8645e2646fce738fe9d8e13fe7d7d947ad4. That is the orogonal commit which fixes this issue in fmt.

cliffburdick:
Is this PR ready for review?

bench/00_misc/fltflt_arithmetic.cu
```diff
@@ -882,8 +882,12 @@ __global__ void iterative_fmod_kernel(T* __restrict__ result, int64_t size, int3
   int64_t idx = blockIdx.x * blockDim.x + threadIdx.x;
   if (idx < size) {
     T val[ILP_FACTOR];
-    T init_val = static_cast<T>(100000.0 * std::numbers::pi);
-    const T divisor = static_cast<T>(std::numbers::e);
+    // std::numbers::pi / std::numbers::e are not available in nvcc device code
```
cliffburdick: I think I'd rather just leave the nvbench code as not testable on the windows platform rather than trying to maintain patches for it since nvbench does not support windows. I think any third-party library not supporting it is a good enough reason not to try to support it.

zcsizmadia:
> Is this PR ready for review?

I think so. I am not changing it. I still think there might some major changes to this PR during the review.  e.g. you mentioned that since nvbench  is not supported, maybe under Windows we should not do nvbench. This would simplify somewhat the PR for sure

bench/00_misc/fltflt_arithmetic.cu
zcsizmadia: I havent seen any major reason that nvbench would not work on windows. In a perfect world, where MatX would use the latest nvbench, with the latest fmt. The windows build of Matx would be possible without any patching

cliffburdick: MSVC isn't listed as supported by nvbench, so we probably don't want to attempt to support it here with patches. I think if they eventually add support we can revisit it.

zcsizmadia: I can remove the patching part of nvbench from the MR. What do you think if I keep the source code changes to the bench cu files (which are very similar in nature to all the other changes). So when nvbench is using the latest fmt, we can just turn on building bench.

cliffburdick: `cuda::std::numbers::pi` is available in device code and we should be able to use it there via CCCL

cliffburdick: I think removing the patch and leaving the CMake MSVC-specific changes is fine. Thanks!

zcsizmadia:
Reverted the nvbench changes. bench cannot be compiled any more, but tests and examples build fine.

include/matx/transforms/fft/fft_cuda.h
```diff
@@ -233,8 +233,8 @@ template <typename OutTensorType, typename InTensorType> class matxCUDAFFTPlan_t
       params.onembed[0] = o.Size(RANK - 1); // Unused
       params.istride = i.Stride(RANK - 1);
       params.ostride = o.Stride(RANK - 1);
-      params.idist = (RANK == 1) ? 1 : i.Stride(RANK - 2);
-      params.odist = (RANK == 1) ? 1 : o.Stride(RANK - 2);
+      params.idist = (RANK == 1) ? 1LL : i.Stride(static_cast<uint32_t>(RANK - 2));
+      params.odist = (RANK == 1) ? 1LL : o.Stride(static_cast<uint32_t>(RANK - 2));
```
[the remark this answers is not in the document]
zcsizmadia: @cliffburdick I think this was inherited from the original code, I am not sure if needs to be addressed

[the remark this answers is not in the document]
zcsizmadia: I could replace that pattern with (?):
```
if constexpr (RANK == 1) {
        params.idist = 1LL;
        params.odist = 1LL;
      } else {
        params.idist = i.Stride(RANK - 2);
        params.odist = o.Stride(RANK - 2);
      }
```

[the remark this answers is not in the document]
cliffburdick: I would ignore this. As the comment says this cannot be hit at runtime so the comment is not relevant.

zcsizmadia:
VS 2026 builds without warning. I am testing VS 2022 in case we need to support that as well. Not sure if we need to go lower than that (VS 2019 has limited c++ 20 support)

zcsizmadia:
Making VS2022 might be harder than it should. Since VS2022 broke many things with std::format,  it is not crazy trivial at first glance. Probably I can make it work theat it is available (std::format) from cpp code, but not from nvcc. I would say that it should be a iteration after this is merged to support VS 2022. I am concerned to break something in the existing code by guarding VS2022.

cliffburdick:
> VS 2026 builds without warning. I am testing VS 2022 in case we need to support that as well. Not sure if we need to go lower than that (VS 2019 has limited c++ 20 support)

I think VS 2026 is a reasonable minimum version given we've had no support before. I also tried 2022 in the past and did not have the amount of time it was going to require to get it working. Let's start with VS2026.

zcsizmadia:
btw, do you have any automation which builds for multiple platforms.? I try to test  locally when I make even a  small change. but I am always concerned that I break something on Linux.

examples/spectrogram_graph.cu
```diff
@@ -57,7 +57,7 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] char **argv)
 
   using complex = cuda::std::complex<float>;
   cudaGraph_t graph;
-  cudaGraphExec_t instance;
+  cudaGraphExec_t instance = nullptr;
```
cliffburdick, line 60: Was this producing an error?

zcsizmadia: Yes. C4700: uninitialized local variable. Which is treated as an error. I was globally hiding certain warnings and chose to fix some. However it is up for discussion as I mentioned in the TODOs. Of course if we globally hide C4700, we dont need this assignment. There are couple of others like that.

zcsizmadia: I listed  the disabled warnings in the PR description

include/matx/core/defines.h
```diff
@@ -104,10 +104,39 @@ namespace matx {
     #define MATX_IGNORE_WARNING_POP_CLANG
 #endif
 
+#if defined(_MSC_VER)
```
cliffburdick: Shouldn't this be in the #elif case above?

zcsizmadia: Makes more sense  that way for sure. I wonder if it will be tweaked, if the MSVC clang compiler needs to be supported, but that is to be seen

include/matx/core/defines.h
```diff
@@ -97,17 +97,43 @@ namespace matx {
...
+    #define MATX_IGNORE_WARNING_PUSH_MSVC(WARN_ID) \
+        __pragma(warning(push)) \
+        __pragma(warning(disable : WARN_ID))
+
+    #define MATX_IGNORE_WARNING_POP_MSVC \
+        __pragma(warning(pop))
+
+    #define MATX_IGNORE_WARNING_NEXT_LINE_MSVC(WARN_ID) \
+        __pragma(warning(suppress : WARN_ID))
 #else
```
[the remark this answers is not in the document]
zcsizmadia: Fixed following the existing pattern

test/00_operators/find_peaks.cu
```diff
@@ -17,9 +17,9 @@ TYPED_TEST(OperatorTestsFloatNonComplexSingleThreadedHostAllExecs, FindPeaks)
   // example-begin findpeaks-test-1
   auto tiv = make_tensor<TestType>({32});
   auto tidx = make_tensor<index_t>({32});
-  tiv.SetVals({ 3.9905e-01, 5.1668e-01, 2.4930e-02, 9.4008e-01, 9.4585e-01, 7.9673e-01, 4.1501e-01, 8.2026e-01, 2.2904e-01, 9.0959e-01, 1.0000e+01,
-    7.5222e-02, 4.0922e-01, 9.6007e-01, 2.0930e-01, 1.9395e-01, 8.9094e-01, 4.3867e-01, 3.5698e-01, 5.4537e-01, 8.2992e-01, 2.0994e-01, 7.6842e-01,
-    4.2899e-01, 2.1167e-01, 6.6055e-01, 1.6536e-01, 4.2499e-01, 9.9267e-01, 6.9642e-01, 2.4719e-01, 7.0281e-01});
+  tiv.SetVals({ static_cast<TestType>(3.9905e-01), static_cast<TestType>(5.1668e-01), static_cast<TestType>(2.4930e-02), static_cast<TestType>(9.4008e-01), static_cast<TestType>(9.4585e-01), static_cast<TestType>(7.9673e-01), static_cast<TestType>(4.1501e-01), static_cast<TestType>(8.2026e-01), static_cast<TestType>(2.2904e-01), static_cast<TestType>(9.0959e-01), static_cast<TestType>(1.0000e+01),
```
cliffburdick, line 20: This is the realm of C++ syntax being prohibitively ugly. Can we change these to C-style casts just to make it easier to read?

test/00_operators/frexp_test.cu
```diff
@@ -29,7 +29,9 @@ TYPED_TEST(OperatorTestsFloatNonComplexNonHalfAllExecs, Frexp)
   exec.sync();
   // Create operators representing fractional and integer
   const auto [ofrac, oint] = frexp(tiv0);
+  MATX_IGNORE_WARNING_PUSH_MSVC(4834) // discarding return value of [[nodiscard]] function (comma operator)
```
cliffburdick, line 32: Why does it think the return value is discarded?

test/00_tensor/Storage.cu
```diff
@@ -130,14 +130,22 @@ public:
     // Duck-typed interface
     void* allocate(size_t bytes) {
         size_t aligned_bytes = (bytes + alignment_ - 1) & ~(alignment_ - 1);
+#ifdef _MSC_VER
+        void* ptr = _aligned_malloc(aligned_bytes, alignment_);
+#else
         void* ptr = std::aligned_alloc(alignment_, aligned_bytes);
```
cliffburdick, line 136: Why does msvc not have this?

zcsizmadia: Microsoft Visual C++ (MSVC) does not support std::aligned_alloc because of a fundamental incompatibility between the C11/C++17 standard requirements and the underlying heap implementation,

https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170

test/00_operators/frexp_test.cu
zcsizmadia: MatX's operator= is [[nodiscard]] — it doesn't execute the assignment eagerly. Instead it returns a lazy set object that must have .run(exec) called on it (MatX's deferred execution model).

MatX also overloads operator, comma.h:232 to capture both operands:

So in (tofrac = ofrac, toint = oint).run(exec):

tofrac = ofrac → [[nodiscard]] object (call it set1)
toint = oint → [[nodiscard]] object (set2)
(set1, set2) → calls MatX's overloaded operator,, producing a CommaOp(set1, set2)
.run(exec) → executes both assignments together
The value of set1 is not actually discarded — it's passed as the left argument to the overloaded operator, and stored in CommaOp.

The warning is a MSVC quirk: C4834 fires whenever a [[nodiscard]] value appears on the left-hand side of a comma expression, before MSVC fully considers that operator, has been overloaded and that the value is actually being consumed. MSVC is being overly conservative here — the suppression is just working around that false positive.

test/00_operators/find_peaks.cu
zcsizmadia: Roger. Started a test build to check it.  Fyi I just followed some existing pattern of static casting constants. But I agree, it does not look pretty :)

test/00_tensor/Storage.cu
zcsizmadia: Fyi, I tried using MSVC clang route (instead of cl). It got very ugly very fast

cliffburdick (approved):

zcsizmadia:
@cliffburdick I rebased my branch with main and some similar errors pop up in the newly committed code. What do you think about the following?

Instead of adding [maybe used]]s or new MSVC pragmas, in this phase 1, all those warnings are silenced by the cmake file. That minimizes the changes in the code, and most importantly, it is way more contributor/user friendly to peeps who make commits. Otherwise they have tow worry about MSVC cl quirks, which is silly. I would document it all the silenced warnings and later phases those can be removed from cmake files and fixed in the code (if needed). The Linux build is the baseline anyway. I really don't want to create headaches for Linux developers with these changes

cliffburdick:
> @cliffburdick I rebased my branch with main and some similar errors pop up in the newly committed code. What do you think about the following?
> 
> Instead of adding [maybe used]]s or new MSVC pragmas, in this phase 1, all those warnings are silenced by the cmake file. That minimizes the changes in the code, and most importantly, it is way more contributor/user friendly to peeps who make commits. Otherwise they have tow worry about MSVC cl quirks, which is silly. I would document it all the silenced warnings and later phases those can be removed from cmake files and fixed in the code (if needed). The Linux build is the baseline anyway. I really don't want to create headaches for Linux developers with these changes

I think that's certainly fine to make msvc ignore the unused variables for now.

cliffburdick:
@zcsizmadia please cherry-pick this commit as well since I needed it: https://github.com/NVIDIA/MatX/commit/54a1c687ed05256ddce75f917e77b8ea4d9bd96f

It successfully compiled on Visual Studio 2026 for me after that,

cliffburdick:
We should also update the main README to remove
```
MatX support is currently limited to Linux only due to the time to test Windows. If you'd like to voice your support for native Windows support using Visual Studio, please comment on the issue here: https://github.com/NVIDIA/MatX/issues/153.
```

And replace with
```
MatX fully supports Linux with an installed CUDA Toolkit. Windows support is currently in alpha and only compiles with Visual Studio 2026. Older Visual Studio releases do not compile. Windows support is currently tested on a best-effort basis and not with every commit.
```

And `For this reason, when running with GPU support, CUDA 12.2.1 and g++9, nvc++ 24.5, or clang 17 or newer is required` should be changed to ` For this reason, when running with GPU support, CUDA 12.2.1 and g++9, nvc++ 24.5, MSVC 14.50, or clang 17 or newer is required`.

We should also update build.rst and have a note about windows support being in alpha/etc.

cliffburdick:
> btw, do you have any automation which builds for multiple platforms.? I try to test locally when I make even a small change. but I am always concerned that I break something on Linux.

We have linux automation in our CI

cliffburdick:
@zcsizmadia are you still working on this one?

cliffburdick:
This PR has been merged with some updates in #1221

PR #1181, opened 2026-05-12
Use Thrust for multithreaded host algorithms

cliffburdick:
The CPU sorting, reduction and prefix sum only used a single-threaded version. This adds the parallel versions via thrust algorithms, Speedups are shown below on a 20-core ARM:

image


cliffburdick:
This passed but CI timed out when making coverage files.

PR #1194, opened 2026-06-01
Add corrmap operator for windowed correlation maps

tbensonatl:
Add the corrmap() operator, which computes 1D or 2D correlation maps. corrmap produces a per-sample or per-pixel measure of correlation between two signals or images (or, generically, tensors) over a small window local to each output sample. The output is the same size as the input. The 1D and 2D corrmap operators are differentiated by the rank of the window lengths provided to the operator. 2D correlation maps are common in SAR. For that use case, use the MAGNITUDE normalization type and take the abs() of the output to get the real-valued correlation values in the range [0, 1].

See the MatX documentation for full details of the corrmap() operator and normalization modes.

### DIFF
--- a/docs_input/api/stats/corr/corrmap.rst
+++ b/docs_input/api/stats/corr/corrmap.rst
@@ -1,0 +1,101 @@
+.. _corrmap_func:
+
+corrmap
+#######
+
+Per-element windowed correlation map between two operators of identical
+shape. Unlike :ref:`corr_func`, which produces a single function-of-lag
+output, ``corrmap`` produces an output of the **same shape as the input**,
+where each output element is a normalized correlation computed over a small
+window of samples around the corresponding input element.
+
+Supports a 1-D window over the last input dimension (signal-processing form)
+and a 2-D window over the last two input dimensions (image-processing form).
+All other dimensions are treated as independent batches. The window is
+cropped at the input boundary: out-of-bounds offsets are skipped and the
+normalization uses only the in-bounds samples (means and energies are
+computed over the cropped window, not over a zero-padded one).
+
+Typical applications:
+
+- SAR / InSAR coherence and interferogram phase (complex inputs, MAGNITUDE normalization)
+- Stereo matching disparity cost (real inputs, ZNCC normalization)
+- Template matching, optical flow, change detection (ZNCC normalization)
+
+.. versionadded:: 1.0.0
+
+.. doxygenfunction:: corrmap(const OpA &A, const OpB &B, index_t window)
+.. doxygenfunction:: corrmap(const OpA &A, const OpB &B, const cuda::std::array<index_t, 2> &window)
+
+Normalization modes
+~~~~~~~~~~~~~~~~~~~
+
+The compile-time template parameter ``Mode`` selects how each window's
+samples are combined. The full mathematical definition of each mode is
+given in its ``CorrMapNormalize`` enum value.
+
+.. doxygenenum:: matx::CorrMapNormalize
+
+Input type requirements
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Inputs must be floating-point or complex floating-point. Supported inner
+scalar types are ``float``, ``double``, ``matx::matxFp16`` /
+``matx::matxBf16`` (and the underlying ``__half`` / ``__nv_bfloat16``),
+and their complex counterparts.
+
+Integer and complex-integer inputs are rejected at compile time by a
+``static_assert``. This applies to every mode, including
+:cpp:enumerator:`matx::CorrMapNormalize::NONE`: although the NONE-mode
+product :math:`A \cdot \bar B` is mathematically well-defined for
+integer operands, the normalized modes (MAGNITUDE / ZNCC) require a
+final division that would silently truncate to zero for integer
+arithmetic, so all integer inputs are rejected uniformly to avoid a
+surprising mode-dependent failure mode. Cast integer inputs explicitly
+before calling, e.g. ``corrmap(as_float(A), as_float(B), w)``.
+
+Output element type
+~~~~~~~~~~~~~~~~~~~
+
+- Complex if either input is complex.
+- Real otherwise.
+- Precision is the greater of the inputs: e.g. ``float + double`` produces
+  ``double``, ``complex<float> + complex<double>`` produces
+  ``complex<double>``, and ``complex<float> + double`` also produces
+  ``complex<double>``.
+
+Window indexing
+~~~~~~~~~~~~~~~
+
+The window uses the floor-center convention: for a
+window of length :math:`w` centered at index :math:`n`, the offsets span
+:math:`[-\lfloor w/2 \rfloor,\ w - 1 - \lfloor w/2 \rfloor]`. Odd window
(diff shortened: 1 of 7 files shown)


include/matx/operators/corrmap.h
```diff
@@ -0,0 +1,832 @@
...
+              "        win_start[d] = center - (window_[d] / 2);\n"
+              "      }}\n"
+              "\n"
+              "      constexpr bool is_cplx = is_complex_v<value_type>;\n"
+              "      constexpr bool need_energies = (Mode == 1);  // MAGNITUDE\n"
+              "      constexpr bool use_welford   = (Mode == 2);  // ZNCC\n"
+              "\n"
+              "      [[maybe_unused]] value_type sum_ab{{}};\n"
+              "      [[maybe_unused]] inner_type sum_aa{{0}};\n"
+              "      [[maybe_unused]] inner_type sum_bb{{0}};\n"
```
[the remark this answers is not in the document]
tbensonatl: Passed the mode selection and all current enum values into JIT string generation in updated commit

include/matx/operators/corrmap.h
```diff
@@ -0,0 +1,832 @@
...
+              "          cuda::std::max(M2_b, inner_type{{0}}));\n"
+              "        if (denom == inner_type{{0}}) return value_type{{}};\n"
+              "        if constexpr (is_cplx) {{\n"
+              "          return value_type{{C.real() / denom, C.imag() / denom}};\n"
+              "        }} else {{\n"
+              "          return C / denom;\n"
+              "        }}\n"
+              "      }}\n"
+              "    }}\n"
+              "  }}\n"
```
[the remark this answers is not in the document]
tbensonatl: Noted. The preferred path is to add support to matx::sqrt() for scalar types and thus remove the need for corrmap_scalar_sqrt entirely.

docs_input/api/stats/corr/corrmap.rst
```diff
@@ -0,0 +1,101 @@
+.. _corrmap_func:
+
+corrmap
```
cliffburdick, line 3: Does this function exist in matlab or python?

include/matx/operators/corrmap.h
```diff
@@ -0,0 +1,837 @@
...
+          is_dynamic_tensor_v<OpA> || is_dynamic_rank_op_v<OpA> ||
+          is_dynamic_tensor_v<OpB> || is_dynamic_rank_op_v<OpB>>;
+
+#ifdef MATX_EN_JIT
+        // JIT storage: same operands the host operator() reads, plus the
+        // runtime window-size array.
+        struct JIT_Storage {
+          typename detail::inner_storage_or_self_t<detail::base_type_t<OpA>> a_;
+          typename detail::inner_storage_or_self_t<detail::base_type_t<OpB>> b_;
+          cuda::std::array<index_t, WindowRank> window_;
```
cliffburdick: Do these need to be in `JIT_Storage` or can they be encoded as compile-time parameters in the emitted function name?

include/matx/operators/corrmap.h
```diff
@@ -0,0 +1,837 @@
...
+
+          for (int32_t i = 0; i < OpA::Rank(); i++) {
+            MATX_ASSERT_STR(a_.Size(i) == b_.Size(i), matxInvalidSize,
+              "corrmap() inputs must have matching shapes in every dimension.");
+            out_dims_[i] = a_.Size(i);
+          }
+        }
+
+        template <typename CapType, typename... Is>
+        __MATX_INLINE__ __MATX_DEVICE__ __MATX_HOST__ decltype(auto) operator()(Is... indices) const
```
cliffburdick: I think the const version should just be `auto` here?

test/00_operators/corrmap_test.cu
```diff
@@ -0,0 +1,649 @@
...
+    const double i = static_cast<double>(d.imag());
+    return std::sqrt(r * r + i * i);
+  } else {
+    return std::abs(static_cast<double>(d));
+  }
+}
+
+// Helper: invoke a generic lambda once per normalization mode (compile-time).
+template <typename F>
+void for_each_mode(F &&fn) {
```
cliffburdick, line 181: Can this be used instead? https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/utility/static_for.html#libcudacxx-extended-api-utility-static-for

cliffburdick (approved):

docs_input/api/stats/corr/corrmap.rst
tbensonatl: I have not been able to find anything identical, but there are related functions. Cross-correlation functions are related (correlate2d() and xcorr2()), but those compute outputs as a function of lag (i.e., a correlation measure between signals where one is shifted). Here, we want the correlation of two source signals/images centered at the same location. It is also similar to template matching functions, but instead of a fixed template, you have a spatially-varying template (i.e., the window from the "other" image).

tbensonatl: Given all that, I'm not sure this is the best name. I didn't find an exact match in well-known libraries, so I chose an otherwise-unused name.

include/matx/operators/corrmap.h
tbensonatl: The window_ values? Yes, I expect those can be compile-time params, potentially with some performance gain. My concern with that is that it may cause more divergence in the JIT vs non-JIT source, so they could more easily drift. I can update the JIT path if we prefer to avoid storage though.

test/00_operators/corrmap_test.cu
tbensonatl: Interesting, I was not aware of that. It looks like we could use it, but we would add the assumption that our CorrMapNormalize scoped enum uses a contiguous range of ints so that we could safely convert the integral_cosntant<> values to an enum. That's true today and will probably remain true, but the current helper just uses the enum values directly.

include/matx/operators/corrmap.h
cliffburdick: Typically any arrays should just be stored as constexpr and encoded into the name of the function so the JIT path has the least overhead possible. slice() is a good example:

https://github.com/NVIDIA/MatX/blob/ae5258391c52591678ecb8f012f54fbac6326206/include/matx/operators/slice.h#L86

https://github.com/NVIDIA/MatX/blob/ae5258391c52591678ecb8f012f54fbac6326206/include/matx/operators/slice.h#L111

Usually this leaves the only thing in the storage struct as the tensors/operators themselves.

include/matx/operators/corrmap.h
tbensonatl: Yes, you're right. I'll update that.

include/matx/operators/corrmap.h
tbensonatl: Okay, I'll update the JIT path.

include/matx/operators/corrmap.h
tbensonatl: Updated

include/matx/operators/corrmap.h
tbensonatl: Added the window as a constexpr array

cliffburdick (approved):

PR #1203, opened 2026-06-29
Fix stack() operator-operand composition

tbensonatl:
Three bugs that only surface when a stack() expression is materialized via run() with an operator (non-tensor) operand:

(A) The stack() factory stored a local `first = pp_get<0>(ts...)` used
    only to call the static Rank() method, so its value was never read.
    Under -Werror=unused-but-set-variable this fails to compile when the
    first operand has a trivial destructor (e.g. any expression operator).

(B) StackOp inherited the no-op BaseOp::PreRun/PostRun and never
    forwarded them to its variadic operands. When an operand is a
    transform operator that allocates temporary storage during PreRun,
    that temporary was never allocated or computed, so operator() read
    uninitialized memory.

(C) StackOp::Size() computed Size(dim-1) on an operand whose Size()
    indexes a fixed-size cuda::std::array. GCC's -Werror=array-bounds
    analysis cannot prove dim-1 >= 0 (axis_ is a runtime int) and
    rejects it.

The existing Stack test only reads operator() elements directly and never materializes a stack expression, so none of these were triggered.

Fixes:
  (A) Call ::Rank() from a type alias for the type of first. The
      alias includes a [[maybe_unused]] because Release builds elide
      the assertion and do not reference the local alias.
  (B) Add PreRun/PostRun to StackOp that forward to every operand via
      cuda::std::apply, guarded by is_matx_op<>().
  (C) Compute the operand dim once and clamp it to a valid non-negative
      index so the fixed-size-array access is provably in-bounds.

Adds StackOperatorInput test to verify Pre/PostRun forwarding.

### DIFF
--- a/include/matx/operators/stack.h
+++ b/include/matx/operators/stack.h
@@ -332,12 +332,17 @@ namespace matx
       {
         if(dim==axis_)
           return sizeof...(Ts);
-        else if (dim < axis_) {
-          return cuda::std::get<0>(ops_).Size(dim);
-        } else {
-          // remove axis_ dim from dim.
-          return cuda::std::get<0>(ops_).Size(dim-1);
+        // Map the requested output dimension to the operand dimension by dropping
+        // the stacked axis. For any valid query op_dim is already >= 0 (a dim > axis_
+        // implies dim >= 1, and the only legitimate dim==0 query is the axis itself,
+        // handled above). The clamp is therefore unreachable for valid input and
+        // exists solely to give the compiler a provable lower bound for
+        // -Werror=array-bounds.
+        int op_dim = (dim < axis_) ? dim : dim - 1;
+        if (op_dim < 0) {
+          op_dim = 0;
         }
+        return cuda::std::get<0>(ops_).Size(op_dim);
       }
 
       __MATX_INLINE__ __MATX_HOST__ int32_t DynRank() const {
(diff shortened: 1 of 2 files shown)


include/matx/operators/stack.h
```diff
@@ -353,6 +358,35 @@ namespace matx
...
+      {
+        cuda::std::apply([&](const auto&... ops) {
+          auto postrun_one = [&](const auto &op) {
+            if constexpr (is_matx_op<cuda::std::decay_t<decltype(op)>>()) {
+              op.PostRun(shape, ex);
+            }
+          };
+          (postrun_one(ops), ...);
+        }, ops_);
+      }
```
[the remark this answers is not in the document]
tbensonatl: Good point - removed [[maybe_unused]] in an updated commit

include/matx/operators/stack.h
```diff
@@ -332,12 +332,17 @@ namespace matx
       {
         if(dim==axis_)
           return sizeof...(Ts);
-        else if (dim < axis_) {
-          return cuda::std::get<0>(ops_).Size(dim);
-        } else {
-          // remove axis_ dim from dim.
-          return cuda::std::get<0>(ops_).Size(dim-1);
+        // Map the requested output dimension to the operand dimension by dropping
```
cliffburdick, line 335: Should we have two conditionals here or just suppress the warning if it can't happen in valid cases?

tbensonatl: The clamping will also map invalid queries like `.Size(-1)` to `.Size(0)`. We don't generally do that elsewhere, so perhaps we should not do it here either. We could just throw instead of setting op_dim to 0. But, without clamping or throwing, the array-bounds error is not necessarily a false positive, so I don't know that we would want to suppress it.

cliffburdick (approved):

PR #1209, opened 2026-07-23
Add streaming filter support for conv1d, resample_poly, channelize_poly

tbensonatl:
Add Conv1DStream, ResamplePolyStream, and ChannelizePolyStream objects that filter arbitrarily long signals delivered in segments. Segments of any size are supplied via feed() and a final flush() emits the trailing outputs and ends the stream. The concatenated outputs match the corresponding one-shot transform. Objects are created with the make_*_stream() factories and support CUDA and host executors.

Each object retains only a small history buffer between calls, and feed()/flush() write into a caller-provided output buffer sized once via max_output(), so no allocation occurs during streaming. conv1d supports FULL/SAME/VALID modes. To support the internal windowed one-shot calls, the resample_poly and channelize_poly transforms and kernels gain output-window offset parameters.

See examples/streaming.cu and the API Reference section Signal and Image Processing / Streaming for usage and per-transform caveats. Tests cover one-shot equivalence across filter/segment configurations, varying segment schedules, end-of-stream semantics, and host executors.


include/matx/streaming/resample_poly_stream.h
```diff
@@ -0,0 +1,460 @@
...
+  template <typename BufOp, typename OutTensor>
+  void resample_exec(const BufOp &buf, index_t lo, index_t cnt, OutTensor &out)
+  {
+    auto os = slice(out, {0}, {cnt});
+    if (ur_ == 1 && dr_ == 1) {
+      // up == down is a special case; just copy the owned inputs directly.
+      // feed() owns the segment's lifecycle around this call, so copy via
+      // exec_.Exec, not run(), to avoid additional segment PreRun/PostRun.
+      auto id_copy = (os = slice(buf, {lo}, {lo + cnt}));
+      exec_.Exec(id_copy);
```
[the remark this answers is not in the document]
tbensonatl: This was a prior design decision made and documented here:https://github.com/NVIDIA/MatX/blob/96ed845af3098dbed7fa90d0d71486dfb1951607/include/matx/transforms/resample_poly.h#L327. The current behavior of treating up=1, down=1 as a no-op matches the behavior of scipy. I agree that applying the filter and treating this as convolution is the more correct approach, but likely we should address that in a separate PR if we wish to change the contract for up=1 down=1.

docs_input/api/signalimage/streaming/channelize_poly_stream.rst
```diff
@@ -0,0 +1,33 @@
+.. _make_channelize_poly_stream_func:
+
+make_channelize_poly_stream
+===========================
+
+Streaming polyphase channelizer: channelize a signal delivered in segments,
+equivalent to a one-shot :ref:`channelize_poly <channelize_poly_func>` over
+the concatenated stream
+
+.. versionadded:: 1.0.0
```
cliffburdick: I think this should be 1.1.0 (the next release)

docs_input/api/signalimage/streaming/index.rst
```diff
@@ -0,0 +1,31 @@
+.. _streaming:
+
+Streaming
+#########
+
+Streaming objects filter arbitrarily long signals delivered in segments. An
```
cliffburdick: Maybe the word process instead of filter here?

docs_input/api/signalimage/streaming/channelize_poly_stream.rst
tbensonatl: Updated

docs_input/api/signalimage/streaming/index.rst
tbensonatl: Updated

tbensonatl:
The initial version of this PR returned a slice of the user-provided output tensor from the feed()/flush() methods. There are cases in which there are zero new output samples, but MatX does not allow for zero-sized slices, so that approach failed in such cases. The API has been updated so that feed()/flush() instead return the count of output samples that have been written to the user-provided output tensor/operator. The user can then slice (if count > 0) or otherwise consume as needed.

cliffburdick (approved):
Nice work!

PR #1224, opened 2026-08-10
[FEA] Match Python's einsum syntax

### LINKED ISSUE #300: [FEA] Match Python's einsum syntax
Our `einsum` syntax is lacking things compared to python:

1) Broadcast notation
2) Optional outputs
3) Single input (no commas)

The first two should be possible without much work. The third needs slightly more parsing to understand what operation is taking place.

MashrafeeAryan:
## What this does

Adds two features to `einsum` that make it work more like NumPy's `einsum`:

1. **Broadcast notation (`...`)**: you can now use `...` to represent batch dimensions, instead of spelling them out with letters. For instance, without broadcast notation, you would do something like: `bcij,bcjk->bcik`. 

However, with broadcast notation you can do following: `...ij,...jk->...ik`.

One limitation is that the batch dimensions represented by `...` must have matching sizes. For example, if `...` represents `[5, 2]` in one tensor, it must also represent `[5, 2]` in the other tensor. MatX does not currently support NumPy-style size-1 broadcasting for these dimensions because cuTensorNet requires shared dimensions to have matching sizes. It's a cuTensorNet limitation. 

2. **Optional output**: you can now leave out `->` and MatX will figure out the output for you. Example: `einsum("ii", a)` works the same as `einsum("ii->", a)`.

3. **Single Input (no comma)**: from my understanding, single input with no comma has already been implemented. For instance, 
`ijkl->jlki` is single input (no comma) and it works. In `test/00_tensor/EinsumTests.cu`, following tests check single input (no comma implementation): Permute, Sum, and Trace. I have not added anything new to it. 

## Testing

Added tests for broadcast notation and optional output.

Ran following commands: 
```
cmake --build build -j32 --target test_00_tensor_EinsumTests
./test_00_tensor_EinsumTests --gtest_filter="*ImplicitOutput*:*Broadcast*"
```

I ran all the tests on:
- Ubuntu 22.04, 
- NVIDIA RTX 4500 Ada Generation GPU, 
- driver version 570.133.07
- CUDA 12.8.

## Docs

I didn’t update the docs in this PR yet. The docs page still says broadcast notation isn't supported. If I get a greenlight, I can go ahead and update the documents.

## Concern
If we are supposed to follow any formatting tools, please let me know. I will fix it immediately. I didn’t find any formatting tools details in the repo.

# Related Issue
Closes #300

### DIFF
--- a/include/matx/transforms/einsum.h
+++ b/include/matx/transforms/einsum.h
@@ -504,7 +504,7 @@ public:
     i = 0;
     ((params.nmodes_[i++] = tensors.Rank()), ...);
 
-    // For tokens with ellipsis, we cannot compare raw string length
+    //For tokens with ellipsis, we cannot compare raw string length
     // to tensor rank. Therefore, ellipsis tokens are validated with EllipsisRank()
 
     i = 0;
@@ -748,9 +748,10 @@ namespace cutensor {
    * MatX uses a syntax very similar to NumPy's einsum syntax:
    * https://numpy.org/doc/stable/reference/generated/numpy.einsum.html
    *
-   * Ellipses are not supported yet, but a variadic list of tensors for contraction is supported. The output
-   * operator '->' is required in MatX currently, and serves to provide error checking on the output tensor size.
-   *
+   * Broadcast ellipses ("...") are supported for aligning batch dimensions across
+   * operands, and a variadic list of tensors for contraction is also supported. The output
+   * operator '->' is optional; if omitted, the output subscript is inferred using NumPy's
+   * implicit-output rule.
    *
    * @tparam OutputType Output tensor type
    * @tparam InT Types of input tensors
--- a/include/matx/transforms/einsum.h
+++ b/include/matx/transforms/einsum.h
@@ -331,7 +331,7 @@ public:
   /**
    * @brief Infers the output subscript for implicit mode einsum without "->"
    * 
-   * If any input token contains a broadcast ellips ("..."), it represents
+   * If any input token contains a broadcast ellipsis ("..."), it represents
    * the other remaining dimensions
    */
   static std::string InferImplicitOutput(const std::vector<std::string> &input_tokens){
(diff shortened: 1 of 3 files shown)


include/matx/transforms/einsum.h
```diff
@@ -318,29 +381,113 @@ class matxEinsumHandle_t {
...
+      matxInvalidDim,
+      "einsum operand rank is too small for the explicit subscripts given alongside \"...\""
+    );
+    return ellipsis_rank;
+  }
+
+  /**
+   * @brief Expands a subscript token into one mode ID per tensor dimension
+   *
+   * "..." is replaced with synthetic mode IDs that align broadcast dimensi
```
cliffburdick: "dimensi"

include/matx/transforms/einsum.h
```diff
@@ -354,20 +501,79 @@ class matxEinsumHandle_t {
...
-    for (i = 0; i < tokens.size(); i++) {
-      for (const char &c: tokens[i]) {
-        params.modes_[i].push_back(static_cast<int32_t>(c));
-      }
-    }
-
     i = 0;
     ((params.nmodes_[i++] = tensors.Rank()), ...);
 
+    //For tokens with ellipsis, we cannot compare raw string length
```
cliffburdick, line 507: Space after //

include/matx/transforms/einsum.h
```diff
@@ -354,20 +501,79 @@ class matxEinsumHandle_t {
...
+    // to tnesor rank. Therefore, ellipsis tokens are validated with EllipsisRank()
```
cliffburdick: "tnesor"

cliffburdick (changes requested):
Just a few minor comments

MashrafeeAryan:
Hi @cliffburdick, thanks for the review! I will fix those comments and push an update shortly. Should I also go ahead and update the documentation in this PR to cover broadcast notation and optional output?

cliffburdick:
> Hi @cliffburdick, thanks for the review! I will fix those comments and push an update shortly. Should I also go ahead and update the documentation in this PR to cover broadcast notation and optional output?

Yes, I missed that. The docs need to be updated and references about "..." not supported can be removed.

MashrafeeAryan:
> > Hi @cliffburdick, thanks for the review! I will fix those comments and push an update shortly. Should I also go ahead and update the documentation in this PR to cover broadcast notation and optional output?
> 
> Yes, I missed that. The docs need to be updated and references about "..." not supported can be removed.

Hi, I have updated the docs and removed references stating that "..." is not supported. I have also addressed the review comments. Thank you for your time!

cliffburdick (approved):

PR #1227, opened 2026-08-17
[FEA] Add support for mdspan

### LINKED ISSUE #582: [FEA] Add support for mdspan
`mdspan` is an upcoming feature in C++23 with two main changes:

1) `operator[]` for multi-dimensional arrays
2) Indexing and types for multi-dimensional arrays

While 1 requires a new C++ compiler, 2 has been included the CCCL recently. For this feature we can add partial support for the second part by taking in `mdspan` types into `make_tensor` calls to translate to what we have.

MashrafeeAryan:
## What this does

Adds two features:

1. Create a MatX tensor from an mdspan

`make_tensor` can now create a MatX tensor from a `cuda::std::mdspan`
 
2. Ability to use `tensor[i, j]` with C++23

When the compiler supports C++23 multidimensional indexing, tensors can use:

```cpp
tensor[2, 3] = value;
auto value = tensor[2, 3];
```
## Testing

Added tests for mdspan and operator[]

Built and ran the tensor tests with:

```bash
cmake --build build \
  --target test_00_tensor_MakeTensorTests \
  --parallel 2

./build/test/test_00_tensor_MakeTensorTests \
  --gtest_filter='MakeTensorTests.*Mdspan*'

./build/test/test_00_tensor_MakeTensorTests
````
Testing was done using:

- Google Colab
- Tesla T4 GPU
- CUDA 12.8.93
- NVIDIA driver 580.82.07
- GCC 11.4.0
- CMake 3.31.10
- C++20

The mdspan changes and the full `MakeTensorTests` suite passed with C++20

The new `operator[]` feature requires C++23 multidimensional subscript support,
which is not available with CUDA 12.8. I could not update the CUDA environment
due to administrative permission restrictions

I would really appreciate it if the C++23 tests could be run in a compatible
environment during review.

## Documentation

Updated the documentation with:
- How to create a MatX tensor from an mdspan
- The new C++23 `tensor[i, j]` syntax
- Continued support for `tensor(i, j)` in C++20 and C++23

## Related issue

Closes #582

### DIFF
--- a/docs_input/api/creation/tensors/make.rst
+++ b/docs_input/api/creation/tensors/make.rst
@@ -38,6 +38,26 @@ Custom Allocator Support
 .. doxygenfunction:: make_tensor( TensorType &tensor, const index_t (&shape)[TensorType::Rank()], Allocator&& alloc)
 .. doxygenfunction:: make_tensor( TensorType &tensor, ShapeType &&shape, Allocator&& alloc)
 
+cuda::std::mdspan Support
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``cuda::std::mdspan`` creates a non-owning tensor using mdspan's existing memory.
+
+.. code-block:: cpp
+
+    #include <cuda/std/mdspan>
+    int data[6]{};
+    using extents_type = cuda::std::extents<matx::index_t, 2, 3>;
+    cuda::std::mdspan<int, extents_type> span{data};
+    auto tensor = matx::make_tensor(span);
+
+The tensor uses the mdspan's original data, dimensions, and strides without
+copying the data. The original data must be valid while the tensor is being used.
+Static and dynamic extents are supported with ``layout_right``,
+``layout_left``, and ``layout_stride``.
+
+.. doxygenfunction:: make_tensor(const cuda::std::mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &span)
+
 DLPack Support
 ~~~~~~~~~~~~~~
 .. versionadded:: 1.1.0
--- a/docs_input/api/creation/tensors/make.rst
+++ b/docs_input/api/creation/tensors/make.rst
@@ -38,25 +38,43 @@ Custom Allocator Support
 .. doxygenfunction:: make_tensor( TensorType &tensor, const index_t (&shape)[TensorType::Rank()], Allocator&& alloc)
 .. doxygenfunction:: make_tensor( TensorType &tensor, ShapeType &&shape, Allocator&& alloc)
 
-cuda::std::mdspan Support
-~~~~~~~~~~~~~~~~~~~~~~~~~
+mdspan Support
+~~~~~~~~~~~~~~
 
-``cuda::std::mdspan`` creates a non-owning tensor using mdspan's existing memory.
+MatX can create non-owning tensors from ``cuda::std::mdspan`` and, when
+available, ``std::mdspan``.
+
+For C++20 and CUDA mdspan:
 
 .. code-block:: cpp
 
     #include <cuda/std/mdspan>
+
     int data[6]{};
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by MashrafeeAryan** [comment]
Hi @cliffburdick, I addressed the issue Greptile pointed out. Please let me know if I need to change anything else. Thank you for your time!

File: docs_input/api/creation/tensors/make.rst [hunk 1]
```diff
@@ -38,6 +38,26 @@ Custom Allocator Support
 .. doxygenfunction:: make_tensor( TensorType &tensor, const index_t (&shape)[TensorType::Rank()], Allocator&& alloc)
 .. doxygenfunction:: make_tensor( TensorType &tensor, ShapeType &&shape, Allocator&& alloc)
 
+cuda::std::mdspan Support
```
cliffburdick: I think we should support both `cuda::std::mdspan` for those on older c++ versions, then an include guard for the same ones for `std::mdspan`. Maybe there's a type trait you can use to simplify the code.

include/matx/core/make_tensor.h
```diff
@@ -620,6 +621,87 @@ auto make_tensor( T* const data,
...
+template <typename ElementType,
+          typename Extents,
+          typename LayoutPolicy,
+          typename AccessorPolicy>
+  // Ensures users can only use layout_right, layout_left, and layout_stride
+  requires(
+    std::is_same_v<LayoutPolicy, cuda::std::layout_right> ||
+    std::is_same_v<LayoutPolicy, cuda::std::layout_left> ||
+    std::is_same_v<LayoutPolicy, cuda::std::layout_stride>)
+auto make_tensor(
```
cliffburdick: I think if we want to support `std::mdspan` also we should add a type trait that's true for `cuda::std::mdspan` and `std::mdspan`. Have that be the only `requires` clause, then do separate checks inside? It could be a bit ugly but at least it keeps it in one function

MashrafeeAryan: Sounds like a good idea.  I am working on it!

MashrafeeAryan: Hi @cliffburdick , I have updated the code to support std::mdspan while using a type trait.

cliffburdick (approved):

include/matx/core/make_tensor.h
```diff
@@ -43,8 +44,75 @@
...
 #include "matx/core/log.h"
+
+#if __has_include(<mdspan>)
+  #include <mdspan>
+#endif
 namespace matx {
 
+namespace detail {
+
+template <typename T>
```
cliffburdick: These should be defined in type_traits.h

cliffburdick (changes requested):
One minor change

include/matx/core/make_tensor.h
MashrafeeAryan: I will define them in type_traits file. However, I just wanted to confirm did you mean `include/matx/core/type_utils.h`? I couldn't find a `type_traits.h` file in the repository.

cliffburdick: Yes, that's the correct file. Sorry for the confusion.

MashrafeeAryan: Thank you so much! I have moved it to type_utils.h.

include/matx/core/make_tensor.h
```diff
@@ -44,74 +43,10 @@
 #include "matx/core/tensor_desc.h"
 #include "matx/core/dlpack.h"
 #include "matx/core/log.h"
+#include "matx/core/type_utils.h"
```
cliffburdick: I think this is already indirectly included by several of the headers above.

MashrafeeAryan: I have removed that and double checked that it compiles without any issue.

cliffburdick (approved):

PR #1230, opened 2026-08-23
Add bulk SAR mocomp operator

tbensonatl:
Add an experimental operator to apply bulk motion compensation to frequency domain SAR phase history history. It applies a frequency-dependent phase ramp to each pulse, referencing its phase to the specified per-pulse reference range.

Add a --bulk-mocomp option to the sarbp example that applies bulk mocomp based on the reference ranges contained in the CPHD file (or in the sarbp input file, which is typically generated from a CPHD). Applying mocomp to data that has already been mocomped will effectively reference the phase to the wrong point (i.e., mocomp is not idempotent and applying it multiple times will generate errors). Most CPHD files are already mocomped, so this option is for special cases where the data is not mocomped or the sarbp input file was generated with non-mocomped data.

### DIFF
--- a/examples/sarbp/sarbp.cu
+++ b/examples/sarbp/sarbp.cu
@@ -1417,7 +1417,7 @@ int main(int argc, char **argv) {
   std::cout << "Pixel-z mode     : " << pixel_z_arg << std::endl;
   std::cout << "Bulk mocomp      : ";
   if (apply_bulk_mocomp) {
-    std::cout << "enabled (double phase arithmetic)";
+    std::cout << "enabled";
   } else {
     std::cout << "disabled";
   }
--- a/examples/sarbp/sarbp.cu
+++ b/examples/sarbp/sarbp.cu
@@ -238,6 +238,7 @@ struct BpRunCtx {
   bool is_fx_domain;
   bool is_int16_mode;
   bool apply_window;
+  bool apply_bulk_mocomp;
   bool taylor_fast_add_third_order;
   SarBpPixelZMode pixel_z_mode;  // compile-time pixel-z assumption to apply
   int sgn;
@@ -246,6 +247,7 @@ struct BpRunCtx {
   std::string gold_file;
   std::string cmap_file;
   double del_r;
+  matx::experimental::SarBulkMocompParams bulk_mocomp_params;
   SarBpParams params;
   double3 *h_positions;
   double *h_range_to_mcp;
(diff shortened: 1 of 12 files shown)


examples/sarbp/sarbp.cu
```diff
@@ -423,29 +423,90 @@ static int run_bp_device(PosTensor blk_positions, RtmTensor blk_rtm,
...
+    make_tensor(
+        blk_compressed, {ctx.block_size, ctx.fft_size},
+        matx::MATX_DEVICE_MEMORY);
+    make_tensor(
+        blk_fx, {ctx.block_size, ctx.num_samples_raw},
+        matx::MATX_DEVICE_MEMORY);
+  }
+
+  // int16 interleaved I/Q samples and per-pulse scale (int16 mode only)
+  tensor_t<int16_t, 2> blk_fx_i16;
```
cliffburdick, line 440: Do these need `[[maybe_unused]]` in case the branch doesn't take it?

test/00_operators/sar_bulk_mocomp_test.cu
```diff
@@ -0,0 +1,513 @@
...
+  using RangeType = cuda::std::tuple_element_t<0, TypeParam>;
+  using ExecType = cuda::std::tuple_element_t<1, TypeParam>;
+
+  ExecType exec{};
+  RunRank2ReferenceCase<RangeType>(exec, 7, -1);
+  RunRank2ReferenceCase<RangeType>(exec, 8, 1);
+  MATX_EXIT_HANDLER();
+}
+
+#ifdef MATX_EN_JIT
```
cliffburdick: Why does the JIT version need a different test instead of just adding the executor?

cliffburdick (approved):

examples/sarbp/sarbp.cu
tbensonatl: Their use is a runtime decision based on reading a flag from an input file. So the compiler at least will not deem it unused, although it may be unused in practice. This convention replaced the previous convention that we always created a tensor, but make it size-1 in the case that we wouldn't use it.

test/00_operators/sar_bulk_mocomp_test.cu
tbensonatl: The other tests do already add the JIT executor when MATX_EN_JIT is defined. This test was verifying that we obtain fusion with block-reduction operators (i.e., sum in `sar_bulk_mocomp(fx, sum(range_components, {1}), params)`). I can just run this same test through the other executors though. It will materialize the sum output to a temporary buffer in those cases.

cliffburdick: How does it verify it? Is it checking that it's a single kernel somewhere?

tbensonatl: No, it doesn't check that there was only a single kernel. It just validates the final output, so it is an indirect test. There would have been no PreRun for the JIT executor, so the temporary buffer would have never been created. Getting correct results at least implies that there was a reduction without the use of temporary output for sum(). I'm not sure how we would verify it directly. Maybe via CUPTI? I updated the test file to run this test under all executors, so the test is no longer JIT-only.

cliffburdick: I don't think we need to verify that. I was just curious.

